PR #10, opened 2020-11-25
Add FPT and move logic from syft to SyMPC repo

gmuraru:


src/sympc/protocol/beaver/beaver.py
```diff
@@ -14,24 +14,33 @@ def build_triples(x, y, op_str):
     if op_str not in EXPECTED_OPS:
         raise ValueError(f"{op_str} should be in {EXPECTED_OPS}")
 
-    session = x.session
+    from sympc.tensor import AdditiveSharingTensor
+    from sympc.tensor import FixedPrecisionTensor
```
LaRiffle, line 18: maybe avoid imports inside functions. 
When we couldn't do this, sometimes we could import let's say `syft` and then call `syft. FixedPrecisionTensor`

src/sympc/protocol/beaver/beaver.py
```diff
@@ -14,24 +14,33 @@ def build_triples(x, y, op_str):
...
+
     shape_x = x.shape
     shape_y = y.shape
+
+    session = x.session
     conf = session.config
```
LaRiffle, line 24: `conf` might not be needed as a  var

src/sympc/protocol/spdz/spdz.py
```diff
@@ -45,17 +46,22 @@ def mul_master(x, y, op_str):
     return shares
 
 
-""" Functions that are executed at a party """
-
+""" Functions that are executed at each party """
 def mul_parties(session, a_share, b_share, c_share, eps, delta, op_str):
```
LaRiffle, line 50: can you add typing info?

src/sympc/session/session.py
```diff
@@ -29,10 +36,54 @@ def get_copy(self):
...
+        session_copy.przs_generators = self.przs_generators
         session_copy.rank = self.rank
         session_copy.session_ptr = [s_ptr for s_ptr in self.session_ptr]
 
         return session_copy
 
-    def setup_mpc(self):
-        raise NotImplementedError("Need to be able to send the session to other parties")
+    def przs_generate_random_elem(self, shape, generators):
+        from ..tensor import FixedPrecisionTensor
```
LaRiffle, line 46: same, try to avoid imports  inside a function

src/sympc/tensor/additive_shared.py
```diff
@@ -18,20 +18,59 @@ def __init__(self, secret=None, shares=None, session=None):
...
+
         elif shares is not None:
             self.shares = shares
 
+
+    @staticmethod
+    def sanity_checks(secret, shape, session):
+        is_remote_secret = False
+
+        if "Pointer" in type(secret).__name__:
```
LaRiffle, line 47: #HackAlert ;)

src/sympc/tensor/additive_shared.py
```diff
@@ -90,54 +129,28 @@ def __apply_private_op(self, y, op_str):
             from ..protocol import spdz
```
LaRiffle, line 129: same import question

src/sympc/tensor/fixed_precision.py
```diff
@@ -0,0 +1,129 @@
+from ..encoder import FixedPointEncoder
+from .utils import modulo
+from ..config import Config
+import operator
```
LaRiffle, line 4: What is operator?

src/sympc/tensor/utils.py
```diff
@@ -1,16 +1,21 @@
...
+        mask_pos = mask_pos.long()
+        x = x - (mask_pos * ring_size)
+        mask_pos = x > max_value
 
-    result = (x + mask * ring_size).long()
+    mask_neg = x < min_value
+    while mask_neg.any():
+        mask_neg = mask_neg.long()
+        x = x + (mask_neg * ring_size)
+        mask_neg = x < min_value
```
LaRiffle, line 19: This might be very costly for AST. We should discuss this

LaRiffle (changes requested):
Great 1st version! I have added questions to first have a better understanding of all the choices you made :)

src/sympc/session/session.py
```diff
@@ -1,38 +1,134 @@
...
+            party.python.List([None, None]) for party in session.parties
+        ]
+
+        for rank in range(nr_parties):
+
+            # TODO: Need to change this with something secure - see CSPRNG
+            seed = random.randint(-(2 ** 31), 2 ** 31)
+            next_rank = (rank + 1) % nr_parties
+
+            gen_current = session.parties[rank].torch.Generator()
```
gmuraru, line 127: Those are sent at the parties

src/sympc/tensor/additive_shared.py
```diff
@@ -1,172 +0,0 @@
-import torch
```
gmuraru, line 1: The AddtiveSharingTensor becomes the ShareTensorCC(ShareTensorControlCenter -- it has a field with pointers to shares)

src/sympc/protocol/beaver/beaver.py
gmuraru: Need to handle this

src/sympc/tensor/additive_shared.py
gmuraru: Yep - I need to move this into a new function

src/sympc/protocol/spdz/spdz.py
```diff
@@ -33,8 +40,12 @@ def mul_master(x, y, op_str):
     eps_plaintext = eps.reconstruct(decode=False)
     delta_plaintext = delta.reconstruct(decode=False)
 
-    with ThreadPoolExecutor(max_workers=nr_parties) as executor:
-        args = list(zip(session.session_ptr, a_sh.shares, b_sh.shares, c_sh.shares))
+    with ThreadPoolExecutor(
+        max_workers=nr_parties, thread_name_prefix="spdz_mul_master"
+    ) as executor:
+        args = list(
+            zip(session.session_ptr, a_sh.share_ptr, b_sh.share_ptr, c_sh.share_ptr)
```
LaRiffle, line 47: share_ptr should maybe be share_ptrS to dispell any confusion?

src/sympc/protocol/spdz/spdz.py
```diff
@@ -33,8 +40,12 @@ def mul_master(x, y, op_str):
     eps_plaintext = eps.reconstruct(decode=False)
     delta_plaintext = delta.reconstruct(decode=False)
 
-    with ThreadPoolExecutor(max_workers=nr_parties) as executor:
-        args = list(zip(session.session_ptr, a_sh.shares, b_sh.shares, c_sh.shares))
+    with ThreadPoolExecutor(
+        max_workers=nr_parties, thread_name_prefix="spdz_mul_master"
+    ) as executor:
```
LaRiffle, line 45: Idea for later: if we package this in a decorator we might end up with a super sexy syntax

src/sympc/protocol/spdz/spdz.py
```diff
@@ -50,18 +61,33 @@ def mul_master(x, y, op_str):
     return shares
 
 
-""" Functions that are executed at a party """
+""" Functions that are executed at each party that holds shares """
+
 
+def mul_parties(
```
LaRiffle, line 67: Maybe we should isolate the code running on shares and the one running on ShareCC? Maybe again this is thoughts for later.

src/sympc/session/session.py
```diff
@@ -1,40 +1,133 @@
...
+        from sympc.tensor import ShareTensor
+
+        gen0, gen1 = generators
+
+        current_share = torch.randint(
+            low=self.min_value,
+            high=self.max_value,
+            size=shape,
+            dtype=self.tensor_type,
+            generator=gen0,
```
LaRiffle, line 86: same remark as above about csprng

src/sympc/session/utils.py
```diff
@@ -0,0 +1,12 @@
+from functools import lru_cache
+import torch
+
+RING_SIZE_TO_TYPE = {2 ** 16: torch.int16, 2 ** 32: torch.int32, 2 ** 64: torch.int64}
```
LaRiffle, line 4: 👍

src/sympc/tensor/share_control.py
```diff
@@ -0,0 +1,249 @@
...
+
+        min_value = session.min_value
+        max_value = session.max_value
+
+        shape = secret.shape
+        tensor_type = session.tensor_type
+
+        random_shares = []
+        for _ in range(nr_parties - 1):
+            rand_value = torch.randint(min_value, max_value, shape, dtype=tensor_type)
```
LaRiffle, line 121: same point about csprng

src/sympc/tensor/share_control.py
```diff
@@ -0,0 +1,249 @@
...
+                share = random_shares[i]
+            elif i < nr_parties - 1:
+                share = random_shares[i] - random_shares[i - 1]
+            else:
+                share = secret - random_shares[i - 1]
+
+            shares.append(share)
+        return shares
+
+    def reconstruct(self, decode=True):
```
LaRiffle, line 140: I love this function

src/sympc/tensor/share_control.py
```diff
@@ -0,0 +1,249 @@
...
+                op(*share_tuple) for share_tuple in zip(self.share_ptr, y.share_ptr)
+            ]
+
+        return result
+
+    def __apply_public_op(self, y, op_str):
+        op = getattr(operator, op_str)
+        if op_str in {"mul"}:
+            shares = [op(share, y) for share in self.share_ptr]
+        else:
```
LaRiffle, line 217: do you really want to isolate 'mul' ?

src/sympc/tensor/share_control.py
```diff
@@ -0,0 +1,249 @@
...
+            operands_shares = [y] + [0 for _ in range(len(self.share_ptr) - 1)]
+            shares = [
+                op(*tuple_shares)
+                for tuple_shares in zip(self.share_ptr, operands_shares)
+            ]
```
LaRiffle, line 222: what does this do?

src/sympc/tensor/share_control.py
```diff
@@ -0,0 +1,249 @@
...
+
+        result = ShareTensorCC(shares=shares, session=self.session)
+        return result
+
+    def __apply_op(self, y, op):
```
LaRiffle, line 227: I guess we will need a strategy for ops with >2 arguments, and to correctly access their .share_ptr attribute?

src/sympc/tensor/utils.py
```diff
@@ -1,14 +1,7 @@
...
-    mask_neg = x < min_val
-    mask = -mask_pos.long() + mask_neg.long()
-
-    result = (x + mask * ring_size).long()
-
-    return result
+def ispointer(obj: Any) -> bool:
+    if type(obj).__name__.endswith("Pointer") and hasattr(obj, "id_at_location"):
+        return True
+    return False
```
LaRiffle, line 7: maybe we can add smthg similar for checking if machine are virtual?

LaRiffle (changes requested):
Great job, I have put a few questions.
Overall I think our strategy would be to put as much as possible at the shareTensor level to have the orchestrator be less and less important (and maybe disappear??)
Also, maybe we can find a better name for ShareTensorCC =) Any ideas?

gmuraru:
> Great job, I have put a few questions.
> Overall I think our strategy would be to put as much as possible at the shareTensor level to have the orchestrator be less and less important (and maybe disappear??)
> Also, maybe we can find a better name for ShareTensorCC =) Any ideas?

Yep - only that for the moment - in Syft - there is the "duet" and one party needs to control what happens there.

Example:
Alice opens a duet client, Bob's connect to it.
All data get's "stored" in the "duet" - hence from here the ```ShareTensorCC``` -- but yep - I think a better name is in order for this -- We need to think of one.

@arturomf94 @aanurraj do you have any ideas? :D

src/sympc/tensor/share_control.py
gmuraru: This was meant to ```add```/```substract``` the ```y``` only from one party - the others will do an operation with 0 -- but I think we can make only the first party do the operation.

src/sympc/tensor/share_control.py
gmuraru: Nope :( -- it was left from the refactoring

src/sympc/tensor/share_control.py
gmuraru: Yep - I opened an issue for it - I thought it would be harder to add the "CSPRNG", but it seems easy peasy - will add it here.

src/sympc/session/utils.py
gmuraru: Will also add the other rings

src/sympc/protocol/spdz/spdz.py
gmuraru: Really like this idea - will do it here.

src/sympc/tensor/share_control.py
gmuraru: Hmm..for the moment I thought only for add, sub, mul. I think we can extend it after this PR?

src/sympc/session/session.py
```diff
@@ -1,40 +1,142 @@
...
+        self.min_value = -(ring_size) // 2
+        self.max_value = (ring_size - 1) // 2
+
+    def przs_generate_random_share(
+        self, shape: Union[tuple, torch.Size], generators: List[torch.Generator]
+    ) -> "ShareTensor":
+
+        from sympc.tensor import ShareTensor
+
+        gen0, gen1 = generators
```
gmuraru, line 83: Those generators are not that secure because we kinda use a "seed" value to initialize them (that seed is - for the moment - not even generated by them).

In a real scenario, I (let's say party 1) will generate a seed and send it to you Theo, and you will do the same with a seed you generated.

src/sympc/session/utils.py
LaRiffle: yeah good idea!

src/sympc/tensor/share_control.py
LaRiffle: Yeah sounds good!

src/sympc/protocol/spdz/spdz.py
```diff
@@ -10,15 +24,12 @@
...
-    """
-    [c] = [a * b]
-    [eps] = [x] - [a]
-    [delta] = [y] - [b]
+def mul_master(x: MPCTensor, y: MPCTensor, op_str: str) -> List[ShareTensor]:
+    """Function that is executed by the orchestrator to multiply two secret values
 
-    Open eps and delta
-    [result] = [c] + eps * [b] + delta * [a] + eps * delta
+    :return: a new set of shares that represents the multiplication
```
LaRiffle, line 30: Shouldn't we use google style?

src/sympc/protocol/spdz/spdz.py
```diff
@@ -28,40 +39,65 @@ def mul_master(x, y, op_str):
...
-
-    shares = [f.result() for f in futures]
+    args = list(
+        map(
+            list,
+            zip(
+                session.session_ptrs, a_sh.share_ptrs, b_sh.share_ptrs, c_sh.share_ptrs
+            ),
+        )
+    )
```
LaRiffle, line 54: Haha this is a bit ugly! But maybe ok for now

src/sympc/utils/utils.py
```diff
@@ -0,0 +1,123 @@
...
+        nr_parties = len(args)
+
+        if args is None:
+            args = [[] for i in range(nr_parties)]
+
+        if kwargs is None:
+            kwargs = {}
+
+        if parties:
+            func_name = f"{fn.__module__}.{fn.__qualname__}"
```
LaRiffle, line 102: small warning here, some PyTorch function had in the past at least pbs with those attributes, so it might fail sometime randomly (but the error will be easy to understand)

LaRiffle (approved):
A few last comments, but it's good for me i'd say! 🚀 

Just, is there a way to avoid force pushing? I'm happy to help if you have troubles when pushing, but I think we should avoid force push, it make re-reviews a bit complicated I need to see agin all the changes!

src/sympc/protocol/spdz/spdz.py
gmuraru: Made it cleaner

src/sympc/utils/utils.py
gmuraru: Hmm..I think we will get an error for this - and we can cross the road at that point

src/sympc/protocol/spdz/spdz.py
gmuraru: This is what I have seen in Syft :\

aanurraj (approved):

PR #31, opened 2020-12-19
Add support for MatMul

LaRiffle:
## Description
There were some fixed to make to spdz (in the order of operations + some hidden bugs due to using ShareTensor and not Tensor)

**Note** the 2 first commits are a mistake, they have already been merged in a previous PR

### DIFF
--- a/src/sympc/protocol/beaver/beaver.py
+++ b/src/sympc/protocol/beaver/beaver.py
@@ -38,7 +38,7 @@ def build_triples(
     )
 
     b = ShareTensor(session=session)
-    b.tensor = torch.empty(size=shape_x, dtype=torch.long).random_(
+    b.tensor = torch.empty(size=shape_y, dtype=torch.long).random_(
         generator=ttp_generator
     )
 
--- a/src/sympc/protocol/beaver/beaver.py
+++ b/src/sympc/protocol/beaver/beaver.py
@@ -2,14 +2,13 @@
 The Beaver Triples
 """
 
+import operator
 from typing import Tuple
 
 import torch
 import torchcsprng as csprng  # type: ignore
-import operator
 
-from sympc.tensor import MPCTensor
-from sympc.tensor import ShareTensor
+from sympc.tensor import MPCTensor, ShareTensor
 
 EXPECTED_OPS = {"matmul", "mul"}
 
--- a/src/sympc/protocol/spdz/spdz.py
+++ b/src/sympc/protocol/spdz/spdz.py
@@ -5,18 +5,15 @@ Contains functions that are run at:
 * the parties that hold the shares
 """
 
+import operator
 from typing import List
 
 import torch
-import operator
-from sympc.tensor import ShareTensor
 
-from sympc.session import Session
 from sympc.protocol.beaver import beaver
-from sympc.tensor import ShareTensor
-from sympc.tensor import MPCTensor
+from sympc.session import Session
+from sympc.tensor import MPCTensor, ShareTensor
 from sympc.utils import parallel_execution
-
 
 EXPECTED_OPS = {"mul", "matmul"}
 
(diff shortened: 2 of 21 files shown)


src/sympc/protocol/spdz/spdz.py
```diff
@@ -81,18 +81,22 @@ def mul_parties(
 
     op = getattr(operator, op_str)
 
-    eps_b = op(eps, b_share)
-    delta_a = op(delta, a_share)
+    eps_b = op(eps, b_share.tensor)
```
LaRiffle, line 84: this was a bit tricky to spot -> `op(eps, b_share)` would trigger `rmatmul` so   `op(b_share,eps)` which doesn't work for matmul because of shape mismatch

src/sympc/protocol/spdz/spdz.py
```diff
@@ -81,18 +81,22 @@ def mul_parties(
...
-    share.tensor //= scale
+
+    share_tensor //= scale
 
     # Convert to our tensor type
-    share.tensor = share.tensor.type(session.tensor_type)
+    share_tensor = share_tensor.type(session.tensor_type)
+
+    share = ShareTensor(session=session)
+    share.tensor = share_tensor
```
LaRiffle, line 100: Again this was tricky to spot the re-encoding if I provided the data in `ShareTensor(data=...)`

src/sympc/tensor/share_tensor.py
```diff
@@ -91,7 +91,9 @@ def sanity_checks(
         """
         if op_str == "mul" and isinstance(y, (float, torch.FloatTensor)):
             y = ShareTensor(data=y, session=x.session)
-        elif op_str in {"add", "sub", "lt", "gt"} and not isinstance(y, ShareTensor):
+        elif op_str in {"add", "sub", "lt", "gt", "matmul"} and not isinstance(
```
LaRiffle, line 94: I'm not sure of this, especially because I'm unsure of what `if op_str == "mul" and isinstance(y, (float, torch.FloatTensor)):` really does

src/sympc/protocol/beaver/beaver.py
```diff
@@ -37,7 +37,7 @@ def build_triples(
     )
 
     b = ShareTensor(session=session)
-    b.tensor = torch.empty(size=shape_x, dtype=torch.long).random_(
+    b.tensor = torch.empty(size=shape_y, dtype=torch.long).random_(
```
gmuraru, line 40: Nice catch! :D

src/sympc/tensor/share_tensor.py
gmuraru: Here I was thinking about the following scenario: if we multiply and is a ```LongTensor``` then we should not convert it, but I think this adds more complexity (as you pointer out here :D)

```
if not isinstance(y, ShareTensor):
   y = ShareTensor
```

What do you think, @LaRiffle ?

gmuraru (approved):

LaRiffle:
Hmmm @gmuraru I think it deserves maybe a complete functionality, to handle cases when people put LongTensor or even other types that we'd like to be compatible?

gmuraru:
> Hmmm @gmuraru I think it deserves maybe a complete functionality, to handle cases when people put LongTensor or even other types that we'd like to be compatible?

Hmmm - I think yes. But for the moment do you think it is ok to keep it like that - I can create a subsequent ticket for it.

LaRiffle:
> > Hmmm @gmuraru I think it deserves maybe a complete functionality, to handle cases when people put LongTensor or even other types that we'd like to be compatible?
> 
> Hmmm - I think yes. But for the moment do you think it is ok to keep it like that - I can create a subsequent ticket for it.

Thanks @gmuraru!

PR #37, opened 2020-12-22
Add isort configuration file

arturomf94:
## Description
It seems there are a couple of potential [conflicts between black and isort](https://black.readthedocs.io/en/stable/compatible_configs.html#isort). Having the configuration in the `setup.cfg` is [one way](https://black.readthedocs.io/en/stable/compatible_configs.html#formats) we can resolve these conflicts easily.

## How has this been tested?
I tested this with our new pre-commit hook and my own editor.

## Checklist
- [ ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)


gmuraru:
Can't we do it like how they have it in [PySyft](https://github.com/OpenMined/PySyft/blob/dev/setup.cfg#L171)?

arturomf94:
Yes! We could. Good point actually! I'll change it.

arturomf94:
> Can't we do it like how they have it in [PySyft](https://github.com/OpenMined/PySyft/blob/dev/setup.cfg#L171)?

Done

setup.cfg
```diff
@@ -113,3 +113,11 @@ exclude =
 # This will be used when updating. Do not change!
 version = 3.2.3
 package = sympc
+
+[isort]
+multi_line_output = 3
```
gmuraru, line 118: NITPICK: Shouldn't we also have ```profile=black``` ? (I am not sure if this is selecting black as the profiler)

Also, maybe it is worth to have this:
```
sections=FUTURE,STDLIB,THIRDPARTY,SYFT,LOCALFOLDER
```
I think this ^^ will put an order on importing stuff?

Also, they have this part.

```
import_heading_future=future
  import_heading_stdlib=stdlib
  import_heading_thirdparty=third party
  import_heading_syft=syft absolute
  import_heading_localfolder=syft relative
```
^^ I am not 100% sure if one depends on the other - but I think it would be nice to have an order on the order of imported libs :D

arturomf94: As far as I understand... We don't have to have the profile **also** (see [this](https://pycqa.github.io/isort/docs/configuration/profiles/)). It's just another way of adding the same thing. But I set the configuration explicitly to be able to tune it as we want. 

Regarding the order.... That's a good point. I'll make the change so as to roughly follow that order too.

arturomf94: I've changed the setup so as to have pretty much the same configuration as syft.

gmuraru (approved):
LGTM!

PR #38, opened 2020-12-23
Madhava/fix torch 1.6.0

madhavajay:
## Description
This PR lowers the torch requirements to torch==1.6.0.

## Affected Dependencies
None

## How has this been tested?
Tests were failing on python 3.6 and now they pass.
CI for all three python versions has been added.

## Checklist

### DIFF
--- a/src/sympc/utils/utils.py
+++ b/src/sympc/utils/utils.py
@@ -10,7 +10,6 @@ from concurrent.futures import ThreadPoolExecutor
 import functools
 from itertools import repeat
 import operator
-import sys
 from typing import Any
 from typing import Callable
 from typing import Dict
@@ -111,11 +110,8 @@ def parallel_execution(
 
         ex_kwargs: Dict[str, Any] = {}
         ex_kwargs["max_workers"] = nr_parties
-
-        # disable additional args in python 3.6
-        if sys.version_info >= (3, 7):
-            ex_kwargs["initializer"] = initializer
-            ex_kwargs["initargs"] = (loop,)
+        ex_kwargs["initializer"] = initializer
+        ex_kwargs["initargs"] = (loop,)
 
         with Executor(**ex_kwargs) as executor:
             for i in range(nr_parties):
--- a/src/sympc/utils/utils.py
+++ b/src/sympc/utils/utils.py
@@ -6,6 +6,7 @@ int any module
 import asyncio
 import functools
 import operator
+import sys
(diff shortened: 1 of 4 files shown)


src/sympc/utils/utils.py
```diff
@@ -100,9 +101,15 @@ def wrapper(
...
 
-        with Executor(
-            max_workers=nr_parties, initializer=initializer, initargs=(loop,)
-        ) as executor:
+        ex_kwargs: Dict[str, Any] = {}
+        ex_kwargs["max_workers"] = nr_parties
+
+        # disable additional args in python 3.6
+        if sys.version_info >= (3, 7):
+            ex_kwargs["initializer"] = initializer
```
gmuraru, line 109: Hmm, and if we have python3.6 I think the Executor would not work? (because it needs that main event loop?)

madhavajay:
I think python 3.6 is possible: https://idolstarastronomer.com/two-futures.html

Code would look something like this but I don't have much asyncio experience so I will park this for the moment.
```
if sys.version_info >= (3, 7):
    ex_kwargs["initializer"] = initializer
    ex_kwargs["initargs"] = (loop,)

    with Executor(**ex_kwargs) as executor:
        for i in range(nr_parties):
            _args = args[i]
            _kwargs = kwargs.get(i, {})
            futures.append(executor.submit(funcs[i], *_args, **_kwargs))

else:
    with Executor(**ex_kwargs) as executor:
        for i in range(nr_parties):
            _args = args[i]
            _kwargs = kwargs.get(i, {})
            futures.append(
                await loop.run_in_executor(executor, funcs[i], *_args, *_kwargs)
            )
```

src/sympc/utils/utils.py
madhavajay: You are right, i think its possible, see comment but I will leave it for the time being and re-address later if we have time.

LaRiffle (approved):

gmuraru:
@madhavajay does this work? (it should be tested with the notebooks from the SMPC folder from Syft)

madhavajay:
Yes, the latest 0.4 branch with this branch and torch 1.6.0 works on the `examples/secure-multi-party-computation/Duet/1-DS-1-DO` notebooks.
```
$ pip freeze | grep torch
torch==1.6.0
torchcsprng==0.1.2
torchvision==0.7.0
```
Screen Shot 2021-01-06 at 2 29 05 pm

PR #70, opened 2021-03-05
Add support for Function Secret Sharing

LaRiffle:
## Description
We need to implement ReLU, ie to have comparison protocols. FSS is a good candidate as most of the logic is already written in a 3rd party lib (sycret)

## How has this been tested?
_Ongoing_

## Roadmap
- [x] Add first version of FSS for comparison between 2 MPCTensors
- [x] Fix cryptostore to support FSS + SPDZ
- [x] Add support for equality
- [x] Add support for =, <, > between a MPCTensors and a public value (int or Tensor)
- [x] Make sure we have strong testing


LaRiffle:
⚠️ @gmuraru  I can't point to your branch on your repo! Maybe you could import it here?

Issue #72, opened 2021-03-06
Add "*.txt" files as a path to trigger the GH workflow


gmuraru:
## Description
Currently, if we change only the ```requirements``` files we would not trigger the tests.
Add the ```**.txt``` as a ```path``` in the GH workflow actions. 

## Are you interested in working on this improvement yourself?
- Yes, I am.

## Additional Context
This will allow us to test different versions of the package we use directly in the CI.

RaghavPrabhakar66:
Hi, I would love to contribute to this issue.

RaghavPrabhakar66:
@gmuraru there is already `requirements*.txt` added to path in tutorials.yml, so is there a need to add `**.txt` to that?

gmuraru:
Yep - we should also add it to the ```tests``` requirement.

gmuraru:
Solved

Issue #76, opened 2021-03-10
Clean dependency files


gmuraru:
## Description
- Remove all the dependencies that we have in the ```requirements.txt``` file and put those in the ```setup.cfg```.
- Remove the ```requirements.txt``` file from the GA workflows. 

## Are you interested in working on this improvement yourself?
- Yes, I am.

## Additional Context
Add any other context or screenshots.

shubhank-saxena:
I would like to take this on @rasswanth-s .
Does this issue also need to shift `requirements.dev.txt` to be shifted as well, since the end goal is to delete `requirements.txt`?

rasswanth-s:
You would have to shift both into `setup.cfg` , you could refer PySyft's  [setup.cfg](https://github.com/OpenMined/PySyft/blob/dev/packages/syft/setup.cfg) .This would involve modifying the workflows files on install ,to be `pip install -e .[dev]` ,  if we modify the setup.cfg , and also the README.

PR #83, opened 2021-03-12
Add "*.txt" files in path for GH workflow.

RaghavPrabhakar66:
## Description
Added "*.txt" files as a path in tests.yml to trigger the GH workflow.

Fixes issue #72.

## Affected Dependencies
None

## How has this been tested?
None

## Checklist
- [x ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,8 @@ on:
       - "**.py"
       - "setup.cfg"
       - ".github/workflows/**.yml"
-      - "**.txt"
+      - "requirements_dev.txt"
+      - "requirements.txt"
 
   pull_request:
     types: [opened, synchronize, reopened]
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,7 @@ on:
       - "**.py"
       - "setup.cfg"
       - ".github/workflows/**.yml"
+      - "**.txt"
 
   pull_request:
     types: [opened, synchronize, reopened]
@@ -15,6 +16,7 @@ on:
       - "**.py"
       - "setup.cfg"
       - ".github/workflows/**.yml"
+      - "**.txt"
 
 jobs:
   python-tests:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,8 @@ on:
       - "**.py"
       - "setup.cfg"
       - ".github/workflows/**.yml"
-      - "**.txt"
+      - "requirements_dev.txt"
+      - "requirements.txt"
 
 jobs:
   python-tests:


arturomf94:
Is this to trigger the workflow whenever there are changes in the requirments?

arturomf94 (changes requested):
I think this is great but maybe it's better to make the file more explicit: 

i.e.

Have `requirements_dev.txt` / `requirements.txt` instead of `**.txt`.

Because I don't think we want to trigger the workflow if we change anything in the license, for instance.

arturomf94:
> I think this is great but maybe it's better to make the file more explicit:
> 
> i.e.
> 
> Have `requirements_dev.txt` / `requirements.txt` instead of `**.txt`.
> 
> Because I don't think we want to trigger the workflow if we change anything in the license, for instance.

@gmuraru what do you think about this?

RaghavPrabhakar66:
@gmuraru  arturomf94 I have another doubt, Issue #76 states that we remove all dependencies from `requirements.txt` and put them to `setup.cfg` and to remove `requirements.txt` from GA workflows. Isn't that issue in contradiction to this one?

arturomf94:
> @gmuraru @arturomf94 I have another doubt, Issue #76 states that we remove all dependencies from `requirements.txt` and put them to `setup.cfg` and to remove `requirements.txt` from GA workflows. Isn't that issue in contradiction to this one?

That's a good point! 🤔 @gmuraru?

gmuraru:
There is also another requirement file named ```requirements_dev.txt```.
Ideally, we would leave only the ```setup.cfg``` needed for installing sympc and the ```requirements_dev.txt``` would be used to install the devs stuff.

arturomf94:
@RaghavPrabhakar66 - I think, for now, let's add both files explicitly: requirements_dev.txt and requirements.txt :+1:

RaghavPrabhakar66:
@gmuraru @arturomf94 I have both requirements.txt to GA path.

.github/workflows/tests.yml
```diff
@@ -8,13 +8,16 @@ on:
...
+      - "requirements_dev.txt"
+      - "requirements.txt"
 
   pull_request:
     types: [opened, synchronize, reopened]
     paths:
       - "**.py"
       - "setup.cfg"
       - ".github/workflows/**.yml"
+      - "**.txt"
```
gmuraru, line 20: Q: Could you also change it here?

RaghavPrabhakar66: Done.

arturomf94:
Thank you @RaghavPrabhakar66. This looks great IMO. Unfortunately, the tests are currently failing but I believe it has nothing to do with your changes. I think it's best if we try to wait until these problems are fixed anyway and we could merge after that.

arturomf94:
Hm... That's really weird. I don't know what could be going on here. @gmuraru - any ideas?

gmuraru:
> Hm... That's really weird. I don't know what could be going on here. @gmuraru - any ideas?

I re-run the tests now to see if it something that happens only sometimes (we should open a ticket if that is the case)

gmuraru (approved):
LGTM! We need to keep an eye on that tolerance tests

arturomf94 (approved):
:)

PR #85, opened 2021-03-15
Add convolution 2d custom layer

gmuraru:
## Description
Add the 2d Conv layer.

## How has this been tested?
- Added test for inference and reconstruction

## Checklist


src/sympc/module/nn/conv.py
```diff
@@ -0,0 +1,120 @@
...
+        module = torch.nn.Conv2d(
+            in_channels=in_channels,
+            out_channels=out_channels,
+            kernel_size=kernel_size_w,
+            bias=bias,
+        )
+        return module
+
+    @staticmethod
+    def eq_close(
```
LaRiffle, line 100: Maybe this could be a method of a parent shared class between Linear and Conv2d -> a nice GFI?

src/sympc/tensor/mpc_tensor.py
```diff
@@ -426,7 +426,7 @@ def conv2d(
         convolution = self.__apply_op(weight, "conv2d", kwargs_=kwargs)
 
         if bias:
-            return convolution + bias
+            return convolution + bias.unsqueeze(1).unsqueeze(1)
```
LaRiffle, line 429: Are you sure the double unsqueeze is needed?

src/sympc/module/nn/conv.py
```diff
@@ -0,0 +1,120 @@
...
+        "kernel_size",
+        "stride",
+        "padding",
+        "dilation",
+        "groups",
+    ]
+
+    in_channels: int
+    out_channels: int
+    kernel_size: Tuple[int, ...]
```
LaRiffle, line 29: Why are these defined but not set as `self.`, contrary to the next `stride`, `padding`... ?

src/sympc/tensor/mpc_tensor.py
gmuraru: There was a problem here because the bias was not broadcasted accordingly :(
Bias was having the shape [10] and I think the convolution was something like [10, X, X] where X was a number.

src/sympc/module/nn/conv.py
gmuraru: Hmm...I have seen that you could specify the type of the attributes, without them being bound to an instance like this (even if some of them are not used).
Should I remove them? (I do not have any strong opinion regarding this, I set those because I have seen them defined like this [in torch](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/conv.py#L56))

LaRiffle: I don't know to be honest! :)

src/sympc/tensor/mpc_tensor.py
LaRiffle: Hmmm that's strange! But ok let's keep it as such for now

LaRiffle (approved):
Looks good to me!

arturomf94 (approved):

PR #89, opened 2021-03-17
Add metaclass to forward tensor specific methods

gmuraru:
## Description
This metaclass it is used to forward torch tensor specific methods/properties to down the shares/tensors.

## Affected Dependencies

## How has this been tested?
- All the tests should pass

## Checklist

### DIFF
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -746,7 +746,7 @@ class MPCTensor(metaclass=SyMPCTensor):
         """Hook a framework method.
 
         Ex:
-         * if we call "numel" we want to forward it only to one share an return
+         * if we call "numel" we want to forward it only to one share and return
         the result (without wrapping it in an MPCShare)
          * if we call "unsqueeze" we want to call it on all the underlying shares
         and we want to wrap those shares in a new MPCTensor
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -30,8 +30,7 @@ METHODS_FORWARD_ALL_SHARES = {}
 
 
 class MPCTensor(metaclass=SyMPCTensor):
-    """This class is used by an orchestrator that wants to do computation on
-    data it does not see.
+    """Used by the orchestrator to compute on the shares.
 
     Arguments:
         session (Session): the session
@@ -67,7 +66,7 @@ class MPCTensor(metaclass=SyMPCTensor):
         shape: Optional[Union[torch.Size, List[int], Tuple[int, ...]]] = None,
         shares: Optional[List[ShareTensor]] = None,
     ) -> None:
-        """Initializer for the MPCTensor. It can be used in two ways:
+        """Initializer for the MPCTensor. It can be used in two ways.
 
         ShareTensorControlCenter can be used in two ways:
         - secret is known by the orchestrator.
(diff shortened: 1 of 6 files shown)


src/sympc/tensor/mpc_tensor.py
```diff
@@ -48,15 +54,18 @@ class MPCTensor:
 
     __slots__ = {"share_ptrs", "session", "shape"}
 
+    # Used by the SyMPCTensor metaclass
+    METHODS_FORWARD = {"numel"}
+    PROPERTIES_FORWARD = {"T"}
```
LaRiffle, line 59: Why is there some methods or functions defined in the ShareTensor dicts ` METHODS_FORWARD` and  `PROPERTIES_FORWARD` that are not in these ones?

LaRiffle: What is the difference in terms of usage between theses dicts (like  `PROPERTIES_FORWARD`) and the one above (like `PROPERTIES_FORWARD_ALL_SHARES`)?

LaRiffle: I'm a bit confused that all this logic is not directly in the metaclass `SyMPCTensor`. I would rather control everything at this level, and given the class that implements the metaclass, I would add either the properties/methods of the ShareTensor or the ones of the MPCTensors (which might actually live in their respective class files)

src/sympc/tensor/mpc_tensor.py
```diff
@@ -624,14 +633,89 @@ def __check_or_convert(value, session) -> "MPCTensor":
...
+        treat it given that we call it.
+
+        Ex:
+         * if we call "shape" we want to call it on the underlying share
+        and return the result
+         * if we call "T" we want to call it on all the underlying shares
+        and wrap the result in an MPCTensor
+
+        Args:
+            property_nam (str): property to hook
```
LaRiffle, line 648: small typo (also in ShareTensor)

LaRiffle (commented):
Just some questions on the code organization!

src/sympc/tensor/mpc_tensor.py
gmuraru: 1. Depending on if it is a method or property we need to properly "hook" it in ```def hook_property``` or ```def hook_method```.
2. There are some methods for which we can get the result only by forwarding the method to only one share tensor - like ```numel```. But there are methods for which we need to send the ```method``` to all the shares, ```transpose```.
3. I added the forwarding mechanism and how it should be handled at a tensor level (not in the metaclass `SyMPCTensor`) because they all have different behaviors (see how `ShareTensor` vs `MPCTensor` deal with the hooking)

src/sympc/tensor/mpc_tensor.py
```diff
@@ -30,12 +30,11 @@
...
 
 class MPCTensor(metaclass=SyMPCTensor):
-    """This class is used by an orchestrator that wants to do computation on
-    data it does not see.
+    """Used by the orchestrator to compute on the shares.
 
     Arguments:
         session (Session): the session
-        secret (Optional[Union[torch.Tensor, float, int]): in case the secret is
+        secret (Optional[sion[torch.Tensor, float, int]): in case the secret is
```
LaRiffle, line 37: typo here?

LaRiffle (approved):
All good! :)

src/sympc/tensor/mpc_tensor.py
gmuraru: yes...:\

src/sympc/tensor/mpc_tensor.py
```diff
@@ -678,24 +702,87 @@ def __check_or_convert(value, session) -> "MPCTensor":
...
+            res = property(property_share_getter, None)
+
+        return res
+
+    @staticmethod
+    def hook_method(method_name: str) -> Callable[..., Any]:
+        """Hook a framework method.
+
+        Ex:
+         * if we call "numel" we want to forward it only to one share an return
```
aanurraj, line 749: TYPO:  one share and* return

PR #98, opened 2021-03-26
Fix sympc tests

gmuraru:
## Description
Get a root client for each party that is part of the communication.

## Affected Dependencies

## How has this been tested?
- The tests should run without any problem

## Checklist


src/sympc/module/nn/conv.py
```diff
@@ -88,7 +88,7 @@ def share_state_dict(
         bias = None
         if ispointer(state_dict):
             weight = state_dict["weight"].resolve_pointer_type()
-            if "bias" in state_dict.keys().get():
+            if "bias" in weight.client.python.List(state_dict).get():
```
aanurraj, line 91: why we are doing this ?

gmuraru: We are looking or the "bias" term in the ```state_dict``` keys (if it is there, we know that we also have the bias)
`state_dict.keys().get()` returns `NotImplemented` (from the SyftCore side).
What we do here --> we create a remote list on the other side (client side) with the keys from the dictionary.

Ex:
```
list({1:3, 4:5}) --> [1,4]
```
and we `get` that list to check if it has `bias` as a key

aanurraj (approved):

LaRiffle (approved):

PR #99, opened 2021-03-26
Replace torch.allclose in tests with np.allclose

### LINKED ISSUE #93: Replace torch.allclose from tests
## Description
We want to be framework agnostic (even if currently we are highly coupled with `torch`).
For this we should remove all `torch` specific functions from the files where there should be "generic"-like code.

One of those functions is ```allclose``` from `torch`.
This issue is to reove the `allclose` function from the test files.

## Are you interested in working on this improvement yourself?
- Yes, I am.

## Additional Context
Check if `pytest` has an ```allclose``` function or check what other libraries (more general than `torch`) are doing this.

linamnt:
## Description
Resolves #93 . One thing to note is that in some cases, we need to run `tensor.detach().numpy()` or else you get the error: `Can't call numpy() on Tensor that requires grad.  Use tensor.detach().numpy() instead.`

## Affected Dependencies
Numpy is required for these tests.  

## How has this been tested?
- I ran the tests again and they all pass. 

## Checklist


LaRiffle (approved):
Thanks a lot! :)

PR #132, opened 2021-04-29
Style/flake8 configuration

### LINKED ISSUE #14: Check warnings flake
## Description
There is a big list of warning which we ignore and we might need some of those

danielorihuela:
## Description
Improve the flake8 configuration of SyMPC project. Closes #14.

## Checklist
- [ ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A) No, I cannot put labels.


danielorihuela:
Before you commented that we could have our flake8 configuration, I already started working @gmuraru.

Essentially I put all the configuratin in the setup.cfg, I think is better to have all the flake8 configuration in one place than divided in two files. Furthermore, I modified the ignored errors (explanation can be seen down below and in the first commit message).

I am not so sure if we want to remove the entire exclude sections. For instance, I decided to exclude all rst, txt and md files since they are not code. Also, I excluded the setup.cfg because it has a lot of fall positive errors. I do not know what you think. You still think is best to remove the exclude sections and fix all the problems?

This message is extracted from the explanation in my first commit:
- F401 since in Pysyft you can find a lot of
`# noqa 401` through the code.
- DAR101 and DAR201 to ignore only on tests.
  Seems fair, since arguments and returns should be already commented on
  the source code.
  Also, it could clatter the code explaining fixtures.
- Exclude rst, txt, md files and the setup.cfg, since they are not
source code.

setup.cfg
```diff
@@ -102,12 +102,32 @@ formats = bdist_wheel
...
 [flake8]
 # Some sane defaults for the code style checker flake8
+ignore =
+    N805
+    F401 F811
+    W503
+    C901 # TODO: Complex function, (gmuraru) I think we should fix this
+per-file-ignores =
+    tests/**:DAR101, DAR201
+max-line-length = 100
```
gmuraru, line 112: Q: I think in the black config we have 88? (Could you add the same values for all -- probably the same one that Syft uses)

danielorihuela: That is interesting. PySyft has a 120, as you told me remove the excludes section, I decided to return the max-line-length to the original that was in the .flake8rc configuration (100). Black has an 88 length line. You have the last word into what should I put: 120,100 or 88.

gmuraru: Hmm..let's stick to the 100, could you also update the black?

tests/sympc/tensor/mpc_tensor_test.py
```diff
@@ -365,13 +363,13 @@ def test_share_get_method_parties_exception(get_clients) -> None:
...
     x_secret = torch.Tensor([1.0, 2.0, 5.0])
-    expected_res = x_secret * x_secret
+    _ = x_secret * x_secret
 
     mpc_tensor1 = x_secret.share(parties=clients[:2])
     mpc_tensor2 = x_secret.share(parties=clients[2:])
 
     with pytest.raises(ValueError):
-        res = mpc_tensor1 * mpc_tensor2
+        _ = mpc_tensor1 * mpc_tensor2
```
gmuraru, line 372: Q: Why not ```mpc_tensor` * mpc_tensor2``` (without the `_ =`)

tests/sympc/tensor/mpc_tensor_test.py
```diff
@@ -54,7 +52,7 @@ def test_op_mpc_different_sessions(get_clients) -> None:
     y = MPCTensor(secret=torch.Tensor([1, -2]), session=session_two)
 
     with pytest.raises(ValueError):
-        z = x + y
+        _ = x + y
```
gmuraru, line 55: Also here, the same comment

gmuraru (changes requested):
Thank you for the contributions! There are only some nitpicks

danielorihuela:
I think I have applied all the suggestions.

Thanks for the feedback!!

gmuraru (approved):
LGTM!

PR #144, opened 2021-05-06
[WIP] Feat/decouple sympc tests from pysyft

### LINKED ISSUE #7: Simulate the operations on the Workers
To properly test SyMPC we need to have installed the syft_0.3.0.

Add a method to simulate the parties operations (send and get) such that we might have a simple way to test some of the operations from our repository without installing syft.

danielorihuela:
## Description
Decouple SyMPC tests from PySyft `Module` class.

Related to #7.

## How has this been tested?
- Already existent tests, which now use a different "Module".

## Checklist
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A). No because I cannot

### DIFF
--- a/stubs/module_stub.py
+++ b/stubs/module_stub.py
@@ -43,7 +43,11 @@ class Module:
     """Stub of the Module class that can be found in PySyft."""
 
     def __init__(self, torch_ref):
-        """Call setup method to initialize the basic attributes."""
+        """Call setup method to initialize the basic attributes.
+
+        Arguments:
+            torch_ref: Reference to the torch library.
+        """
         self.setup(torch_ref=torch_ref)
 
     def setup(self, torch_ref):
--- a/benchmarks/__init__.py
+++ b/benchmarks/__init__.py
@@ -1,0 +1,1 @@
+"""Module which contains benchmarks tests."""
--- a/benchmarks/module/conv_model.py
+++ b/benchmarks/module/conv_model.py
@@ -12,7 +12,7 @@ from stubs.module_stub import Module
 from sympc.session import Session
 from sympc.session import SessionManager
 from sympc.tensor import MPCTensor
-from tests.stubs.module_stub import Module
+from stubs.module_stub import Module
 
 
 class ConvNet(Module):
--- a/benchmarks/module/conv_model.py
+++ b/benchmarks/module/conv_model.py
@@ -6,15 +6,15 @@ from typing import Callable
 from typing import List
 
 # third party
-import syft as sy
 import torch
 
 from sympc.session import Session
 from sympc.session import SessionManager
 from sympc.tensor import MPCTensor
+from tests.stubs.module_stub import Module
 
 
-class ConvNet(sy.Module):
+class ConvNet(Module):
     """Simple convolutional network."""
 
     def __init__(self, torch_ref):
--- a/benchmarks/module/conv_model.py
+++ b/benchmarks/module/conv_model.py
@@ -8,6 +8,7 @@ from typing import List
 # third party
 import torch
 
+from stubs.module_stub import Module
 from sympc.session import Session
 from sympc.session import SessionManager
 from sympc.tensor import MPCTensor
--- a/stubs/virtual_machine.py
+++ b/stubs/virtual_machine.py
@@ -1,0 +1,1088 @@
+from random import randint
+import pandas as pd
+
+
+class Serializable:
+    """When we want a custom object to be serializable within the Syft ecosystem
+    (as outline in the tutorial above), the first thing we need to do is have it
+    subclass from this class. You must then do the following in order for the
+    subclass to be properly implemented:
+
+    - implement a protobuf file in the "PySyft/proto" folder for this custom class.
+    - compile the protobuf file by running `bash scripts/build_proto`
+    - find the generated python file in syft.proto
+    - import the generated protobuf class into my custom class
+    - implement get_protobuf_schema
+    - implement <my class>._object2proto() method to serialize the object to protobuf
+    - implement <my class>._proto2object() to deserialize the protobuf object
+
+    At this point, your class should be ready to serialize and deserialize! Don't
+    forget to add tests for your object!
+
+    If you want to wrap an existing type (like a torch.tensor) to be used in our serialization
+    ecosystem, you should consider wrapping it. Wrapping means that we NEVER use the wrapper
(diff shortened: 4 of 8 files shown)


danielorihuela:
I decided to put the Module Stub in a tests/stubs/module_stub. Tests fails because all files must follow the pattern *_test. Where should I put the stub?

gmuraru:
But we still have the "workers" coupling problem, right? (because we use `sy.VirtualMachine`) and the `send` + `get` methods.

With this you decouple only the Module, right?

gmuraru:
Ahh! I talked with @tudorcebere and for having a better `decoupling` (not for the workers) we could also move the `allowlist` in SyMPC, from Syft. creating now an issue for that

gmuraru:
Did you run those changes with the Duet Notebooks from Syft (we have the notebooks in the ```examples``` folder)?

danielorihuela:
1. True, I am going to put this as WIP and keep working.
2. No, I did not run those tests. I only created a Module stub so they should keep working. Anyway, I still have work to do with the VirtualMachines as you said.

Thanks!

danielorihuela:
I double checked on the Duet network MNIST example after doing `pip install .` on the sympc folder. It all keeps working.

Your question: "Did you run those changes with the Duet Notebooks from Syft (we have the notebooks in the examples folder)?". Makes me feel like we are not on the same page. I think that I did not explain myself well enough.
What I am doing at the moment is creating fake classes for the tests. So, SyMPC will still be dependent on the PySyft on its source code. Nevertheless, SyMPC tests will no longer be dependent on PySyft code.

SyMPC (src) will still be dependent on PySyft.
SyMPC (tests) will *NOT* depend on PySyft code anymore.

I think that's the easier way to remove the Circular dependency kind of error. I am aware that this does not decouples SyMPC from PySyft completely. So SyMPC will not be a standalone yet, which I understand that is the end goal. However, doing all that in one step seemed like to much work to be done at once. For that reason I am focusing on decoupling the tests. To make PySyft and SyMPC development easier.

stubs/module_stub.py
```diff
@@ -0,0 +1,208 @@
+"""File containing a stub of the Module that can be found in the PySyft library.
```
gmuraru, line 1: Really like the explanations!

gmuraru:
> I double checked on the Duet network MNIST example after doing `pip install .` on the sympc folder. It all keeps working.
> 
> Your question: "Did you run those changes with the Duet Notebooks from Syft (we have the notebooks in the examples folder)?". Makes me feel like we are not on the same page. I think that I did not explain myself well enough.
> What I am doing at the moment is creating fake classes for the tests. So, SyMPC will still be dependent on the PySyft on its source code. Nevertheless, SyMPC tests will no longer be dependent on PySyft code.
> 
> SyMPC (src) will still be dependent on PySyft.
> SyMPC (tests) will _NOT_ depend on PySyft code anymore.
> 
> I think that's the easier way to remove the Circular dependency kind of error. I am aware that this does not decouples SyMPC from PySyft completely. So SyMPC will not be a standalone yet, which I understand that is the end goal. However, doing all that in one step seemed like to much work to be done at once. For that reason I am focusing on decoupling the tests. To make PySyft and SyMPC development easier.
> 
> If you are okey with that reasoning. Then the only thing left to do would be to "copy" the VirtualMachine into SyMPC and remove all the unwanted code. Building a fake VirtualMachine with the minim amount of code that makes tests work.

Yep! you are right that we should do this step by step! Maybe we can merge this PR after getting the other logic decoupling PRs first.

The question regarding the notebook, in Syft, we have some Duet notebooks specific to SyMPC - my question was more targeted if you tested those notebooks - not the Syft specific ones (we have there a notebook where we do Inference using Duet)

stubs/module_stub.py
danielorihuela: Thanks

danielorihuela:
> > I double checked on the Duet network MNIST example after doing `pip install .` on the sympc folder. It all keeps working.
> > Your question: "Did you run those changes with the Duet Notebooks from Syft (we have the notebooks in the examples folder)?". Makes me feel like we are not on the same page. I think that I did not explain myself well enough.
> > What I am doing at the moment is creating fake classes for the tests. So, SyMPC will still be dependent on the PySyft on its source code. Nevertheless, SyMPC tests will no longer be dependent on PySyft code.
> > SyMPC (src) will still be dependent on PySyft.
> > SyMPC (tests) will _NOT_ depend on PySyft code anymore.
> > I think that's the easier way to remove the Circular dependency kind of error. I am aware that this does not decouples SyMPC from PySyft completely. So SyMPC will not be a standalone yet, which I understand that is the end goal. However, doing all that in one step seemed like to much work to be done at once. For that reason I am focusing on decoupling the tests. To make PySyft and SyMPC development easier.
> > If you are okey with that reasoning. Then the only thing left to do would be to "copy" the VirtualMachine into SyMPC and remove all the unwanted code. Building a fake VirtualMachine with the minim amount of code that makes tests work.
> 
> Yep! you are right that we should do this step by step! Maybe we can merge this PR after getting the other logic decoupling PRs first.
> 
> The question regarding the notebook, in Syft, we have some Duet notebooks specific to SyMPC - my question was more targeted if you tested those notebooks - not the Syft specific ones (we have there a notebook where we do Inference using Duet)

1. Okey, I will continue working on the "decoupling" of the VirtualMachine, that I saw has a lot of code involved due to inheritance and other classes involder. VirtualMachine -> Node -> AbstractNode -> Address -> Serializable. VirtualServerConnection is also involder wihch inherits from ServerConnection. If no new member wants to solve #155 I will take it.
2. PySyft has a lot of notebooks. I cannot seem to find the PySyft notebooks targeting SyMPC.

danielorihuela:
Stubing the VirtualMachine is difficult with my current knowledge of the system. There are to much code involved. Only the VirtualMachine has 5 levels of inheritance (VirtualMachine -> Node -> AbstractNode -> Address -> Serializable). VirtualMachine calls functions that again, involved multiple levels of inheritance.

I partially stubbed the VirtualMachine but I get the following error.
image
image

Furthermore, when trying to implement the VirtualMachine, the tests that worked with the Module stub started failing. I do not have the required knowledge of the whole system.
For that reason I am closing this.

gmuraru:
I think we can "park" this issue for the moment - we need to think at a higher-level general API that should call the lower-level specific API (in our case Syft)

Issue #155, opened 2021-05-09
Move allowlist from Syft to SyMPC


gmuraru:
## What?
Currently, we have a dependency for running tests because we might require some functionality from Syft. One of those is the list of allowed operations we can run on a `ShareTensor` -- they are added in ```alowlist``` in (```src/syft/lib/sympc/__init__.py```).

When we add a new operation to the `SharingTensor` and want to call that operation remotely, we need to add it to that allowlist (this requires 2 PRs in 2 repositories).

Move the allowlist into our repository and make Syft only import it.

## Why?
By moving the allowlist into our repo, we decouple this logic from Syft and SyMPC.

danielorihuela:
This should be closed.

gmuraru:
Done

PR #162, opened 2021-05-10
feat: move allowlist from syft to sympc

danielorihuela:
## Description
Move the allow list of operations over share tensors from pysyft to sympc.
Decouple list of methods from pysyft, giving the control to sympc.

Related to #155.
#155 can be closed when [issue of pysyft](https://github.com/OpenMined/PySyft/pull/5554) is closed too.

## Affected Dependencies
PySyft.

## How has this been tested?
- I executed sympc related tests of pysyft.

I am not sure if I should test something specific.

## Checklist
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A). No because I cannot


src/sympc/__init__.py
```diff
@@ -53,3 +53,64 @@ def add_methods_tensor_syft() -> None:
     del get_distribution, DistributionNotFound
 
 add_methods_tensor_syft()
+
+allowed_operations_on_share_tensor = [
```
gmuraru, line 57: Q: Could you move this in the `tensor` folder and here expose this through the `__all__` variable?

gmuraru (changes requested):
No, I think it makes sense here!

Could you rename this to ```allowed_external_api_share_tensor``` (or if you have a better name? - I think having `allowed_operation_on_share_tensor` might indicate to the user that we allow only `operations` --> which in my head translates to methods). We also allow the creation of remote instances, access attributes, etc. 

Could you only validate this with a test - it should

**LE**: Ahh..now that I see this - could you move the other "allowlist" here? (like the one for classes/modules)

gmuraru:
> > No, I think it makes sense here!
> > Could you rename this to `allowed_external_api_share_tensor` (or if you have a better name? - I think having `allowed_operation_on_share_tensor` might indicate to the user that we allow only `operations` --> which in my head translates to methods). We also allow the creation of remote instances, access attributes, etc.
> > Could you only validate this with a test - it should
> > **LE**: Ahh..now that I see this - could you move the other "allowlist" here? (like the one for classes/modules)
> 
> Okey, then the names would be:
> 
> * `allowed_external_api_share_tensor` (methods)
> * `allowed_external_api_modules` (modules)
> * `allowed_external_api_classes` (classes)
>   Also, should I put the classes and modules list in sympc/**init**.py ?
> 
> I do not have enough knowledge of the system, but I would like to raise a yellow flag. I understand that that this will give SyMPC control and eliminate the need of two PR. However, we should keep in mind some things for the future (maybe create new issues).
> 
> 1. There are some lines of the "operations allowlist" in which syft appear. `        ("sympc.store.CryptoStore.get_primitives_from_store", "syft.lib.python.List"),`. If SyMPC is going to be a standalone in the long run, we should keep that in mind.
> 2. I do not know if makes sense to have lists of allowed modules, classes and methods to be used from the outside. Should not the code itself handle the visibility of the different elements?

If we do this with the allowlist, is only to keep the logic separated between SyMPC and PySyft (what we do here should not impact PySyft and vice-versa -- that's why I was thinking to move "our" stuff from PySyft here).

I totally agree with you that the best scenario would be to don't have those lists, only that for the moment, those lists are used to call remote operations on different types (a thing that PySyft knows how to do it).

We also have chunks of code where we do `client.lib.python.List()` (in SyMPC code) -- we are again coupled with the Pysyft I think -- I am still thinking if we can remove this coupling in a "nice and clean way".

danielorihuela:
> Could you only validate this with a test - it should

What exactly should I test?

gmuraru:
> > Could you only validate this with a test - it should
> 
> What exactly should I test?

Ignore me! there should be no test for this since it would already be covered by our tests

src/sympc/tensor/share_tensor.py
```diff
@@ -439,3 +439,65 @@ def method(
     __rmatmul__ = rmatmul
     __truediv__ = div
     __xor__ = xor
+
+
+allowed_external_api_share_tensor = [
```
gmuraru, line 444: This needs to be split, right?

gmuraru: Maybe we can put those in a folder named `api` and then have 3 lists like:
`
sympc.api.allowed_external_classes
sympc.api.allowed_external_attrs
sympc.api.allowed_external_modules
```
Approximately the same name you have in the Syft module

File: src/sympc/tensor/__init__.py [hunk 3]
```diff
@@ -4,5 +4,11 @@
...
 from .mpc_tensor import METHODS_TO_ADD
 from .mpc_tensor import MPCTensor
+from .share_tensor import allowed_external_api_share_tensor
 
-__all__ = ["ShareTensor", "MPCTensor", "METHODS_TO_ADD"]
+__all__ = [
+    "ShareTensor",
+    "MPCTensor",
+    "METHODS_TO_ADD",
+    "allowed_external_api_share_tensor",
```
gmuraru, line 13: And we don't have any changes to this file

src/sympc/__init__.py
```diff
@@ -53,3 +56,27 @@ def add_methods_tensor_syft() -> None:
     del get_distribution, DistributionNotFound
 
 add_methods_tensor_syft()
+
+
+allowed_external_api_modules = [
```
gmuraru, line 61: Those changes would be moved to ```sympc/tensor/api.py```

src/sympc/api.py
```diff
@@ -0,0 +1,114 @@
...
+"""
+
+import sympc
+
+from . import approximations  # noqa: 401
+from . import config  # noqa: 401
+from . import encoder  # noqa: 401
+from . import protocol  # noqa: 401
+from . import session  # noqa: 401
+from . import store  # noqa: 401
```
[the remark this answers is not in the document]
danielorihuela: Yeah sorry, I need to focus more.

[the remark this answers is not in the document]
gmuraru: No worries! You do an excellent job! :D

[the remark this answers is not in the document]
danielorihuela: Thanks!

gmuraru (approved):
LGTM

PR #178, opened 2021-05-17
POC sharing at protocol level

gmuraru:
## Description
Move the distribution of the share on a protocol basis because each protocol might have a different version of how it implements sharing. Falcon uses RSS and each party holds 2 shares.

## How has this been tested?
- it should already be tested

## Checklist

### DIFF
--- a/src/sympc/protocol/fss/fss.py
+++ b/src/sympc/protocol/fss/fss.py
@@ -180,6 +180,8 @@ def fss_op(x1: MPCTensor, x2: MPCTensor, op="eq") -> MPCTensor:
 class FSS(metaclass=Protocol):
     """Function Secret Sharing."""
 
+    tensor_class = ShareTensor
+
     @staticmethod
     def eq(x1: MPCTensor, x2: MPCTensor) -> MPCTensor:
         """Equal operator.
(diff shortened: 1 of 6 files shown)


kamathhrishi:
I think this abstraction works really well as compared to having different distribute share methods at MPCTensor.

src/sympc/protocol/fss/fss.py
```diff
@@ -180,6 +180,8 @@ def fss_op(x1: MPCTensor, x2: MPCTensor, op="eq") -> MPCTensor:
 class FSS(metaclass=Protocol):
     """Function Secret Sharing."""
```
kamathhrishi, line 181: Add the purpose of the share class to the FSS docstring that describes what it does.

src/sympc/protocol/fss/fss.py
```diff
@@ -180,6 +180,8 @@ def fss_op(x1: MPCTensor, x2: MPCTensor, op="eq") -> MPCTensor:
...
 
+    tensor_class = ShareTensor
```
kamathhrishi, line 183: This should be changed to share_class

kamathhrishi: Also, why is share_class not an attribute of FSS?

gmuraru: What do you mean by attribute?

kamathhrishi: Ahh, never mind. I got the answer to my question. I just realized protocol functions are static.

LaRiffle (approved):
Looks good to me!
Any tests need be changed?

gmuraru:
> Looks good to me!
> Any tests need be changed?

I can add a test to test specifcly for this functionality.

Issue #176, opened 2021-05-16
Add `reduction` param to mse_loss


gmuraru:
## Description
Currently, we default (and only do) the `sum` reduction method. We should also add the average one.

## Are you interested in working on this improvement yourself?
- Yes, I am.
- 
## Additional Context
A test should be added to validate the behavior.

libratiger:
Can you provide more information? so I can help on this issues

gmuraru:
For the moment we default to the `sum` reduction operation when doing the `mse_loss`. Ideally, we would want to have a parameter to select if we want `sum` or `avg` . This is dependent on another PR for training :D

gmuraru:
Is is merged now :D - the function is `mse_loss`

libratiger:
I see. I will work on this issue

gmuraru:
Added you to the issue

libratiger:
I just add one PR, please review it @gmuraru

gmuraru:
This si done in [here](https://github.com/OpenMined/SyMPC/pull/180)

PR #180, opened 2021-05-20
Add the mean reduction for mse_loss

libratiger:
## Description
Add the mean reduction for mse_loss solves issue https://github.com/OpenMined/SyMPC/issues/176

## Affected Dependencies

## How has this been tested?
complete the already exist test file

## Checklist

### DIFF
--- a/src/sympc/module/nn/functional.py
+++ b/src/sympc/module/nn/functional.py
@@ -16,7 +16,9 @@ def relu(x: MPCTensor) -> MPCTensor:
     return res
 
 
-def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPCTensor:
+def mse_loss(pred: MPCTensor,
+             target: MPCTensor,
+             reduction: str = "mean") -> MPCTensor:
     """Mean Squared Error loss.
 
     Args:
@@ -31,7 +33,7 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
         ValueError: If `reduction` not in supported methods
     """
     if reduction == "mean":
-        result = (pred - target).pow(2).sum() * 1.0 / len(pred)
+        result = (pred - target).pow(2).sum() / pred.shape[0]
     elif reduction == "sum":
         result = (pred - target).pow(2).sum()
     else:
--- a/src/sympc/module/nn/functional.py
+++ b/src/sympc/module/nn/functional.py
@@ -16,14 +16,24 @@ def relu(x: MPCTensor) -> MPCTensor:
     return res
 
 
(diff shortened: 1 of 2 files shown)


aanurraj (commented):
Also run pre-commit hook.

tests/sympc/module/nn/functional_test.py
```diff
@@ -34,7 +34,16 @@ def test_mse_loss(get_clients) -> None:
     y_pred = torch.Tensor([0.1, 0.3, 0.4, 0.2])
     y_pred_mpc = MPCTensor(secret=y_pred, session=session)
```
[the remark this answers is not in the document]
libratiger: I will add later by a new commit
```suggestion
    y_pred_mpc = MPCTensor(secret=y_pred, session=session)
```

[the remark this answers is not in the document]
libratiger: I just fix this

libratiger:
> Also run pre-commit hook.

It's Ok now

libratiger:
there is no  `len` method for `torch.tensor`, is there any design consideration. i just open one issue: https://github.com/OpenMined/SyMPC/issues/182

src/sympc/module/nn/functional.py
```diff
@@ -16,14 +16,24 @@ def relu(x: MPCTensor) -> MPCTensor:
...
 
     Returns:
         The loss
+
+    Raises:
+        ValueError: If `reduction` not in supported methods
     """
-    return (pred - target).pow(2).sum()
+    if reduction == "mean":
+        result = (pred - target).pow(2).sum() * 1.0 / len(pred)
```
gmuraru, line 34: Q: Why multiply by 1.0?
Why not use size rather than len?

libratiger: the sum maybe a int, but the mean should be flout

gmuraru: in MPC it should still be an integer

libratiger: I will fix with your advice, but can you explain to me with more detail?

gmuraru: Sure.

We have the `MPCTensor` class which holds tensor to `ShareTensor`s (the secret value is split across multiple parties which holds those `ShareTensors`). `ShareTensor`s are in `FixedPointPrecision` (integers).

When we use this `encoding` with `FixedPointPrecision` we transform numbers like the following:
`123.23232` ---> `12323` (this happens if we use a base of 10 and a precision of 2).

All our values are encoded and they are decoded back when we want to reconstruct the result of the computation.

libratiger: Thanks, I update the fix

aanurraj (approved):
Looks Good 🎉

PR #190, opened 2021-05-23
Use Session UUID for retrieving session in parties

gmuraru:
## Description
Don't use the ```session_ptr``` anymore when calling some remote methods and use the ```session_uuid```.
When the orchestrator sends the session to the other parties, it would generate a UUID for each sent session and the party would save the session locally on their side - Check this [PR](https://github.com/OpenMined/PySyft/pull/5600) in the `session_utils.py`. At this point, each party has the session.

When we want to use ```parallel_computation``` we can pass the session_uuid (is unique for each parry) or use the share_tensors which should have a `session_uuid` attached to them.

## Affected Dependencies
- How `ShareTensor`, `MPCTensor`, and `Session` are used and serialized.

## How has this been tested?
- [ ] The unit tests are passing

## Checklist

### DIFF
--- a/src/sympc/protocol/fss/fss.py
+++ b/src/sympc/protocol/fss/fss.py
@@ -20,13 +20,13 @@ import torch as th
 import torchcsprng as csprng  # type: ignore
 
 from sympc.protocol.protocol import Protocol
+from sympc.session import get_session
 from sympc.store import CryptoPrimitiveProvider
 from sympc.store import register_primitive_generator
 from sympc.store import register_primitive_store_add
 from sympc.store import register_primitive_store_get
 from sympc.tensor import MPCTensor
 from sympc.tensor import ShareTensor
-from sympc.utils import get_session
 from sympc.utils import parallel_execution
 
 ttp_generator = csprng.create_random_device_generator()
--- a/src/sympc/protocol/fss/fss.py
+++ b/src/sympc/protocol/fss/fss.py
@@ -27,7 +27,7 @@ from sympc.store import register_primitive_store_add
 from sympc.store import register_primitive_store_get
 from sympc.tensor import MPCTensor
 from sympc.tensor import ShareTensor
-from sympc.tensor.tensor import SyMPCTensor
+from sympc.utils import get_session
 from sympc.utils import parallel_execution
 
 ttp_generator = csprng.create_random_device_generator()
@@ -153,7 +153,7 @@ def fss_op(x1: MPCTensor, x2: MPCTensor, op="eq") -> MPCTensor:
         p_kwargs={},
     )
 
-    args = zip(session.session_ptrs, x1.share_ptrs, x2.share_ptrs)
+    args = zip(x1.share_ptrs, x2.share_ptrs)
     args = [list(el) + [op] for el in args]
 
     shares = parallel_execution(mask_builder, session.parties)(args)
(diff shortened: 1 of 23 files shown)


src/sympc/protocol/spdz/spdz.py
```diff
@@ -255,14 +256,16 @@ def mul_parties(
...
+    share = ShareTensor(session_uuid=UUID(session_uuid_str), config=session.config)
     share.tensor = share_tensor
 
     # Ideally this should stay in the MPCTensor
     # Step 1. Do spdz_mul
     # Step 2. Divide by scale
     # This is done here to reduce one round of communication
     if session.nr_parties == 2:
+        print(share.fp_encoder.precision)
+        print(share.fp_encoder.scale)
```
aanurraj, line 268: Q: are they left intentionally?

src/sympc/store/crypto_primitive_provider.py
```diff
@@ -50,14 +51,21 @@ def generate_primitives(
         generator = CryptoPrimitiveProvider._func_providers[op_str]
         primitives = generator(**g_kwargs)
 
+        for remote_session_uuid, primitive in zip(
+            session.rank_to_uuid.values(), primitives
+        ):
+            print(remote_session_uuid)
```
aanurraj, line 57: Q:Same here too

src/sympc/tensor/mpc_tensor.py
```diff
@@ -206,6 +199,28 @@ def __init__(
...
+            session (Session): Session for which those shares were generated
+
+        Returns:
+            List of ShareTensorPointers.
+        """
+        rank_to_uuid = session.rank_to_uuid
+        parties = session.parties
+
+        share_ptrs = []
+        for rank, share in enumerate(shares):
```
aanurraj, line 217: Q: What rank signifies here?

src/sympc/protocol/fss/fss.py
```diff
@@ -108,7 +108,12 @@ def evaluate(session: Session, b, x_masked, op, dtype="long") -> ShareTensor:
     dtype_options = {None: th.long, "int": th.int32, "long": th.long}
     result = th.tensor(result_share, dtype=dtype_options[dtype])
 
-    share_result = ShareTensor(data=result, session=session)
+    # stdlib
+    import dataclasses
+
+    share_result = ShareTensor(
+        data=result, session_uuid=session.uuid, **dataclasses.asdict(session.config)
```
linamnt, line 115: I think you do not have to make it a dictionary and unpack the config here, you can pass it to the `config` arg, because the items inside (`encoder_base` and `encoder_precision`) are no longer part of the init.

src/sympc/protocol/spdz/spdz.py
gmuraru: No...they are from debugging :))

src/sympc/tensor/mpc_tensor.py
gmuraru: Each party has a rank starting from 0

src/sympc/protocol/spdz/spdz.py
```diff
@@ -52,30 +53,23 @@ def mul_master(
...
+            "a_shape": shape_x,
+            "b_shape": shape_y,
+            "nr_parties": session.nr_parties,
+            **kwargs_,
+        },
+        p_kwargs={"a_shape": shape_x, "b_shape": shape_y},
+    )
+    mask = parallel_execution(spdz_mask, session.parties)(args)
 
     eps_shares, delta_shares = zip(*mask)
```
aanurraj, line 74: We had this because we were not checking for pre-generated primitive and using the exception as a way to generated the primitive on the fly... will it still work ?

gmuraru: Ahh...you are right, I removed this by mistake (the `try...catch` block)

src/sympc/session/__init__.py
```diff
@@ -1,6 +1,14 @@
...
 
 from .session import Session
 from .session_manager import SessionManager
+from .session_utils import get_session
+from .session_utils import set_session
 
-__all__ = ["Session", "SessionManager"]
+# Mapping for rank -> Session
+# rank is needed in the case we have VirtualMachines since we share the same workspace
+CURRENT_SESSION: Dict[int, Session] = {}
```
rasswanth-s, line 12: Q: The methods in session_util map `uuid string` to Session objects,here the type annotation sspecified rank(int)

src/sympc/session/session_manager.py
```diff
@@ -50,11 +43,19 @@ def setup_mpc(session: Session) -> None:
...
-            session.rank = rank
-            session.session_ptrs.append(session.send(party))  # type: ignore
+            session_party = session.copy()
+            session_party.rank = rank
 
+            # And a new uuid
+            session_party.uuid = uuid4()
+            uuids[rank] = session_party.uuid
+            session.session_ptrs.append(session_party.send(party))  # type: ignore
+
```
rasswanth-s, line 56: Q: The `set_session`  which maps parties `uuid` to Session objects, should it assigned it SessionManager class.

rasswanth-s, line 56: I think `set_session` should be assigned here to map the uuid to specific party's session object.

src/sympc/protocol/spdz/spdz.py
```diff
@@ -52,10 +54,11 @@ def mul_master(
...
     session = x.session
-
     shape_x = tuple(x.shape)
     shape_y = tuple(y.shape)
 
+    args = [list(el) + [op_str] for el in zip(x.share_ptrs, y.share_ptrs)]
+
     args = [
         list(el) + [op_str]
         for el in zip(session.session_ptrs, x.share_ptrs, y.share_ptrs)
```
rasswanth-s, line 64: The same `args` variable is assigned twice here.

src/sympc/tensor/mpc_tensor.py
```diff
@@ -206,6 +199,28 @@ def __init__(
         self.grad_fn = None
         self.parents: List["MPCTensor"] = []
 
+    @staticmethod
+    def distribute_shares(shares: List[ShareTensor], session: Session):
```
rasswanth-s, line 203: Q: George , In your Previous [PR](https://github.com/OpenMined/SyMPC/pull/178),share distribution was made at Protocol level.Now it is modified again to distribute at MPC Tensor?

src/sympc/protocol/spdz/spdz.py
gmuraru: it should be removed

src/sympc/tensor/mpc_tensor.py
gmuraru: I will replace this.

rasswanth-s (approved):
Looks good to me.

PR #202, opened 2021-05-29
Fixed Point Encoding - RSTensor

rasswanth-s:
## Description
Implementation of  fixed point encoding in RSTensor

## Affected Dependencies

## How has this been tested?
Added tests for fixed point encoding.

## Checklist

### DIFF
--- a/src/sympc/tensor/replicatedshare_tensor.py
+++ b/src/sympc/tensor/replicatedshare_tensor.py
@@ -78,15 +78,19 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
         return self.fp_encoder.encode(data)
 
     def decode(self):
-        """Decode via FixedPrecisionEncoder.
+        """Decode via FixedPointEncoder.
 
         Returns:
-            torch.Tensor: Decoded value
+            List[torch.Tensor]: Decoded values
         """
         return self._decode()
 
     def _decode(self):
+        """Decodes shares list of RSTensor via FixedPointEncoder.
 
+        Returns:
+            List[torch.Tensor]: Decoded values
+        """
         shares = []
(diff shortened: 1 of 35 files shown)


src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -41,17 +44,55 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
...
+
+        Returns:
+            torch.Tensor: Decoded value
         """
-        self.session = session
-        self.shares = shares
+        return self._decode()
+
+    def _decode(self):
+
```
gmuraru, line 89: NITPICK: Remove empty space + add docstring?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -41,17 +44,55 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
...
+        shares = []
+        for i in range(len(self.shares)):
+            tensor = self.fp_encoder.decode(self.shares[i].type(torch.LongTensor))
```
gmuraru, line 92: Q: We convert it to long also in `ShareTensor`?

rasswanth-s: Yeah ,George ,the semantics is that ,during decoding in the dividend [step](/home/rasswanth/.local/share/virtualenvs/python39-xmdK_WSC/lib/python3.9) ,the scale is cast to the tensor type for floor division. 

For Ex: if tensor_type is` torch.int16` and the scale is `2**17` , casting scale to 16 bit would result in zero which leads to `ZeroDivisonError`. So we cast to `torch.LongTensor` to accomodate large scale values.

PR #207, opened 2021-05-31
Test for RSTensor send and get. Also added equal operation for RSTensor

kamathhrishi:
## Description
Added test for sending and getting RSTtensor which will pass after merging PR for RSTensor proto definition https://github.com/OpenMined/PySyft/pull/5598. 

## How has this been tested?
Test to check send and get functionality for RSTensor 

## Checklist


src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -129,6 +132,14 @@ def eq(self, y):
 
         """
 
+        if not (torch.cat(self.shares) == torch.cat(y.shares)).all():
```
gmuraru, line 135: Q: Why it is needed to call `torch.cat`? and not simply do `if not (self.shares == y.shares)`

kamathhrishi: It was throwing an error `E       RuntimeError: Boolean value of Tensor with more than one value is ambiguous` so I cast each of them into a torch tensor and performed the operation. self.shares and y.shares are a list of tensors, now they become 2d torch tensors after using cat.

rasswanth-s (approved):
Looks good to me.

rasswanth-s (approved):
Looks good to me.

Issue #164, opened 2021-05-12
Replicated Sharing Tensor Distribution


kamathhrishi:
## Feature Description
Develop functionality for creating and distributing shares with Replicatedsharedtensor. 

## Additional Context
This has a dependency on #145, which initializes stub for RSS.

kamathhrishi:
@gmuraru, I will take this up.

Issue #218, opened 2021-06-05
Move the operations implementation in a DefaultProtcol class


gmuraru:
## Description
As we would (possibly) have more protocols that a user could choose from, we would want to also have a ```DefaultProtocol``` in case the protocol specified by the user does not have the operation implemented.

This would work like this:
```
def mul (self, y) -> MPCTensor:
    mul_protocol = getattr(self.session.protocol.mul, None)
    if mul_protocol:  # look for the operation in the protocol the user specified
       return mul_protocol(self, y)
    return DefaultProtocol.mul(self, y)
```
:arrow_up:  this would keep the `MPCTensor` class more clean.

## Are you interested in working on this improvement yourself?
- Yes, I am.

## Additional Context
This would help us split the logic of the implementation for various operations and keep them in a protocol base class.

danielorihuela:
I am interested in better understanding how SyMPC works, I would like to work on that issue. I am not sure of how to do it, so I welcome anyone that wants to help me.

kamathhrishi:
@danielorihuela Sure Daniel, you can reach out to me on OM Slack channel. My username is kamathhrishi.

PR #220, opened 2021-06-05
Distribute shares and reconstruct functionality for RSTensor 

kamathhrishi:
## Description
This PR addresses #164. It adds functionality to distribute shares and reconstruct for replicatedsharedTensor. In order to accommodate the same the logic for reconstruction is moved to sharetensor and RSTensor. 

The following are yet to be completed:

- [x] Type annotations 
- [x] Proper docstrings with args and return types
- [x] Implement malicious setting reconstruction 
- [x] Toggle reconstruction between semi-honest and malicious setting 
- [x] Add granular test for distribution and reconstruction 
- [x] Add exception test for malicious reconstruction 

## How has this been tested?
- [x] A test for distributing shares and reconstructing shares which is equal to shared secret 
- [x]  Test functionality with various number of parties

## Checklist

### DIFF
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -371,7 +371,7 @@ class MPCTensor(metaclass=SyMPCTensor):
         Returns:
             torch.Tensor. The secret reconstructed.
         """
-        plaintext = self.session.protocol.share_class.reconstruct(self.share_ptrs)
+        plaintext = self.session.protocol.share_class.reconstruct(self.share_ptrs,get_shares=get_shares)
 
         if get_shares:
 
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -371,17 +371,15 @@ class MPCTensor(metaclass=SyMPCTensor):
         Returns:
             torch.Tensor. The secret reconstructed.
         """
-        plaintext = self.session.protocol.share_class.reconstruct(
+        result = self.session.protocol.share_class.reconstruct(
             self.share_ptrs,
             get_shares=get_shares,
             security_type=self.session.protocol.security_type,
         )
 
-        print("MPC plaintext: ", plaintext)
-
         if get_shares:
 
-            return plaintext
+            return result
 
         if decode:
             fp_encoder = FixedPointEncoder(
@@ -389,9 +387,9 @@ class MPCTensor(metaclass=SyMPCTensor):
                 precision=self.session.config.encoder_precision,
             )
 
-            plaintext = fp_encoder.decode(plaintext)
+            result = fp_encoder.decode(result)
 
-        return plaintext
+        return result
 
     get = reconstruct
 
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -356,7 +356,7 @@ class MPCTensor(metaclass=SyMPCTensor):
         return shares
 
     def reconstruct(
-        self, decode: bool = True
+        self, decode: bool = True, get_shares: bool = False
     ) -> Union[torch.Tensor, List[torch.Tensor]]:
         """Reconstruct the secret.
 
@@ -366,11 +366,16 @@ class MPCTensor(metaclass=SyMPCTensor):
 
         Args:
             decode (bool): True if decode using FixedPointEncoder. Defaults to True
+            get_shares (bool): Retrieve only shares.
 
         Returns:
             torch.Tensor. The secret reconstructed.
         """
         plaintext = self.session.protocol.share_class.reconstruct(self.share_ptrs)
+
+        if get_shares:
+
+            return plaintext
 
         if decode:
             fp_encoder = FixedPointEncoder(
@@ -379,10 +384,6 @@ class MPCTensor(metaclass=SyMPCTensor):
             )
 
             plaintext = fp_encoder.decode(plaintext)
-
-        else:
-
-            return plaintext
 
         return plaintext
 
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -372,7 +372,7 @@ class MPCTensor(metaclass=SyMPCTensor):
             torch.Tensor. The secret reconstructed.
         """
         plaintext = self.session.protocol.share_class.reconstruct(
-            self.share_ptrs, get_shares=get_shares
+            self.share_ptrs, self.session.protocol.security_type, get_shares=get_shares
         )
 
         if get_shares:
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -768,6 +768,14 @@ class MPCTensor(metaclass=SyMPCTensor):
 
         return result
 
+    def __len__(self) -> int:
+        """Return the length of MPCTensor.
+
+        Returns:
+            int: Length
+        """
+        return self.shape[0]
+
     def __str__(self) -> str:
         """Return the string representation of MPCTensor.
 
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -822,7 +822,7 @@ class MPCTensor(metaclass=SyMPCTensor):
         return object.__getattribute__(self, attr_name)
 
     def backward(self, gradient: Optional["MPCTensor"] = None) -> None:
-        """Perform the backward step on the computationl graph.
+        """Perform the backward step on the computational graph.
 
         Args:
             gradient (MPCTensor): The gradient (received) from the computational graph
@@ -963,7 +963,7 @@ class MPCTensor(metaclass=SyMPCTensor):
         return res
 
     def le(self, other: "MPCTensor") -> "MPCTensor":
-        """Lower or than operator.
+        """Lower or equal operator.
 
         Args:
             other (MPCTensor): MPCTensor to compare.
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -287,6 +287,7 @@ class MPCTensor(metaclass=SyMPCTensor):
 
         Given a secret, split it into a number of shares such that each
         party would get one.
+
         Args:
             secret (Union[ShareTensor, torch.Tensor, float, int]): Secret to split.
             nr_parties (int): Number of parties to split the scret.
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -372,7 +372,9 @@ class MPCTensor(metaclass=SyMPCTensor):
             torch.Tensor. The secret reconstructed.
         """
         plaintext = self.session.protocol.share_class.reconstruct(
-            self.share_ptrs, self.session.protocol.security_type, get_shares=get_shares
+            self.share_ptrs,
+            get_shares=get_shares,
+            security_type=self.session.protocol.security_type,
         )
 
         if get_shares:
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -371,7 +371,9 @@ class MPCTensor(metaclass=SyMPCTensor):
         Returns:
             torch.Tensor. The secret reconstructed.
         """
-        plaintext = self.session.protocol.share_class.reconstruct(self.share_ptrs,get_shares=get_shares)
+        plaintext = self.session.protocol.share_class.reconstruct(
+            self.share_ptrs, get_shares=get_shares
+        )
 
         if get_shares:
 
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -21,9 +21,7 @@ from sympc.config import Config
 from sympc.encoder import FixedPointEncoder
 from sympc.session import Session
 from sympc.tensor import ShareTensor
-from sympc.utils import islocal
 from sympc.utils import ispointer
-from sympc.utils import parallel_execution
 
 from .tensor import SyMPCTensor
 
@@ -289,7 +287,6 @@ class MPCTensor(metaclass=SyMPCTensor):
 
         Given a secret, split it into a number of shares such that each
         party would get one.
-
         Args:
             secret (Union[ShareTensor, torch.Tensor, float, int]): Secret to split.
             nr_parties (int): Number of parties to split the scret.
@@ -373,42 +370,21 @@ class MPCTensor(metaclass=SyMPCTensor):
         Returns:
             torch.Tensor. The secret reconstructed.
         """
+        plaintext = self.session.protocol.share_class.reconstruct(
+            self.share_ptrs, get_shares
+        )
 
-        def _request_and_get(share_ptr: ShareTensor) -> ShareTensor:
-            """Function used to request and get a share - Duet Setup.
-
-            Args:
-                share_ptr (ShareTensor): a ShareTensor
-
-            Returns:
-                ShareTensor. The ShareTensor in local.
-
-            """
-            if not islocal(share_ptr):
-                share_ptr.request(block=True)
-            res = share_ptr.get_copy()
-            return res
-
-        request = _request_and_get
-        request_wrap = parallel_execution(request)
-
-        args = [[share] for share in self.share_ptrs]
-        local_shares = request_wrap(args)
-
-        shares = [share.tensor for share in local_shares]
-
-        if get_shares:
-            return shares
-
-        plaintext = sum(shares)
-
-        if decode:
+        if decode and not get_shares:
             fp_encoder = FixedPointEncoder(
                 base=self.session.config.encoder_base,
                 precision=self.session.config.encoder_precision,
             )
 
             plaintext = fp_encoder.decode(plaintext)
+
+        else:
+
+            return plaintext
 
         return plaintext
 
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -356,7 +356,7 @@ class MPCTensor(metaclass=SyMPCTensor):
         return shares
 
     def reconstruct(
-        self, decode: bool = True, get_shares: bool = False
+        self, decode: bool = True
     ) -> Union[torch.Tensor, List[torch.Tensor]]:
         """Reconstruct the secret.
 
@@ -366,16 +366,13 @@ class MPCTensor(metaclass=SyMPCTensor):
 
         Args:
             decode (bool): True if decode using FixedPointEncoder. Defaults to True
-            get_shares (bool): True if get shares. Defaults to False.
 
         Returns:
             torch.Tensor. The secret reconstructed.
         """
-        plaintext = self.session.protocol.share_class.reconstruct(
-            self.share_ptrs, get_shares
-        )
+        plaintext = self.session.protocol.share_class.reconstruct(self.share_ptrs)
 
-        if decode and not get_shares:
+        if decode:
             fp_encoder = FixedPointEncoder(
                 base=self.session.config.encoder_base,
                 precision=self.session.config.encoder_precision,
--- a/src/sympc/tensor/mpc_tensor.py
+++ b/src/sympc/tensor/mpc_tensor.py
@@ -377,6 +377,8 @@ class MPCTensor(metaclass=SyMPCTensor):
             security_type=self.session.protocol.security_type,
         )
 
+        print("MPC plaintext: ", plaintext)
+
         if get_shares:
 
             return plaintext
--- a/src/sympc/tensor/replicatedshare_tensor.py
+++ b/src/sympc/tensor/replicatedshare_tensor.py
@@ -266,8 +266,6 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
         shares1 = share_ptrs[0].get_shares()[0].get()
         shares2 = share_ptrs[1].get_shares().get()
 
-        shares = []
-
         shares = [shares1] + shares2
 
         if get_shares:
--- a/src/sympc/tensor/replicatedshare_tensor.py
+++ b/src/sympc/tensor/replicatedshare_tensor.py
@@ -15,6 +15,8 @@ import torch
 
 from sympc.config import Config
 from sympc.encoder import FixedPointEncoder
+from sympc.session import Session
+from sympc.tensor import ShareTensor
 from sympc.utils import get_type_from_ring
 
 from .tensor import SyMPCTensor
@@ -64,9 +66,15 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
         self.ring_size = ring_size
 
         self.config = config
-        self.fp_encoder = FixedPointEncoder(
-            base=config.encoder_base, precision=config.encoder_precision
-        )
+        self.fp_encoder = None
+
+        if (self.config.encoder_precision is not None) and (
+            self.config.encoder_base is not None
+        ):
+
+            self.fp_encoder = FixedPointEncoder(
+                base=config.encoder_base, precision=config.encoder_precision
+            )
 
         tensor_type = get_type_from_ring(ring_size)
         self.shares = []
@@ -75,7 +83,11 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
                 self.shares.append(self._encode(shares[i]).to(tensor_type))
 
     def _encode(self, data):
-        return self.fp_encoder.encode(data)
+
+        if self.fp_encoder:
+            return self.fp_encoder.encode(data)
+
+        return data
 
     def decode(self):
         """Decode via FixedPointEncoder.
@@ -92,11 +104,21 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
             List[torch.Tensor]: Decoded values
         """
         shares = []
-        for i in range(len(self.shares)):
-            tensor = self.fp_encoder.decode(self.shares[i].type(torch.LongTensor))
-            shares.append(tensor)
+
+        if self.fp_encoder:
+            for i in range(len(self.shares)):
+                tensor = self.fp_encoder.decode(self.shares[i].type(torch.LongTensor))
+                shares.append(tensor)
+
+        else:
+
+            return self.shares
 
         return shares
+
+    def get_shares(self):
+        """Get shares."""
+        return self.shares
 
     def add(self, y):
         """Apply the "add" operation between "self" and "y".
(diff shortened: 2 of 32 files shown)


src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +241,62 @@ def ne(self, y):
...
+        Args:
+            share_ptrs (List[RSTensorPointers]): List of RSTensor pointers.
+            get_shares (boolean): Return reconstructed values or just shares.
+
+        Returns:
+            reconstructed_value (torch.Tensor): Reconstructed value.
+        """
+        shares1 = share_ptrs[0].get_shares()[0].get()
+        shares2 = share_ptrs[1].get_shares().get()
+
```
gmuraru, line 257: NITPICK:
```
shares = shares2.append(shares1)
if get_shares:
    return shares

return sum(shares)
```

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -92,12 +110,26 @@ def _decode(self):
...
         shares = []
-        for i in range(len(self.shares)):
-            tensor = self.fp_encoder.decode(self.shares[i].type(torch.LongTensor))
-            shares.append(tensor)
+
+        if self.fp_encoder:
+            for i in range(len(self.shares)):
+                tensor = self.fp_encoder.decode(self.shares[i].type(torch.LongTensor))
+                shares.append(tensor)
+
```
gmuraru, line 118: MAYBE: Remove those empty lines

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -64,9 +67,20 @@ def __init__(
...
 
         self.config = config
-        self.fp_encoder = FixedPointEncoder(
-            base=config.encoder_base, precision=config.encoder_precision
-        )
+        self.fp_encoder = None
+
+        if (
+            self.config.encoder_precision is not None
+            and (not isinstance(self.config.encoder_precision, _SyNone))
```
gmuraru, line 74: OUGHTTO: We should check fi ```None``` here. Not the specific type from Syft

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -137,3 +139,17 @@ def test_hook_property(get_clients) -> None:
 
     assert (rst.T.shares[0] == x.T).all()
     assert (rst.T.shares[1] == y.T).all()
+
+
+def test_rst_distribute_reconstruct(get_clients) -> None:
```
aanurraj, line 144: Q: Can we be more specific with the test like checking the number of shares each party is holding?

src/sympc/tensor/mpc_tensor.py
```diff
@@ -373,43 +371,22 @@ def reconstruct(
...
             fp_encoder = FixedPointEncoder(
                 base=self.session.config.encoder_base,
                 precision=self.session.config.encoder_precision,
             )
 
             plaintext = fp_encoder.decode(plaintext)
 
+        else:
+
+            return plaintext
```
rasswanth-s, line 388: MAYBE : Currently we do not allow if `decode = True` and `get_shares=True` , we only output the shares without decoding even if the parameter is true. Should we allow it?I am not sure of a use case where this would be useful,we might have in the future.

tests/sympc/tensor/replicatedshare_tensor_test.py
kamathhrishi: Sure, that's a good suggestion. Will do that.

src/sympc/tensor/mpc_tensor.py
kamathhrishi: Hmmmm, this was present in the original implementation of sharetensor. Not sure why it was intended. That's a good question. @gmuraru Why does the orchestrator need the original shares instead of just value?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +231,62 @@ def ne(self, y):
...
+
+        ptr_list = []
+        for i in range(0, len(parties)):
+            party_ptrs = []
+
+            for j in range(i, i + nshares):
+                if j < len(parties):
+                    tensor = shares[j].tensor
+                    party_ptrs.append(tensor)
+
```
rasswanth-s, line 277: MAYBE : Hrishi, I think we could use modulo approach ,similar like PRZS code ,to avoid two inner loops.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +231,62 @@ def ne(self, y):
...
+            session (Session): Session.
+
+        Returns:
+            List of ShareTensorPointers.
+
+        """
+        parties = session.parties
+
+        nshares = len(parties) - 1
+
```
rasswanth-s, line 268: I think `session_uuid` should also mapped to created RSTensor,similar like what we did for ShareTensor.

src/sympc/tensor/mpc_tensor.py
gmuraru: The `get_shares` and `decode` it is used in the spdz protocol.
Maybe we should add a disclaimer to the function saying that ```get_shares``` has priority over `decode`.

In SPDZ `decode` is used when we reconstruct the `eps` and `delta` values because the MPC tensor has attached a session with config, but we do not intend to decode those values.
The `get_shares` --- hmmm. I think it could be removed. and in the SPDZ we could replace:
```
      z_shares_local = z.get_shares()
      common_args = [z_shares_local, y]
```
with
```
      common_args = [z.share_ptrs, y]
```

Could you try to see if it works?

kamathhrishi: I think SPDZ doesn't use it anymore. https://github.com/OpenMined/SyMPC/blob/5144a24779a72778c22a2a539b2b5e1aca4f2991/src/sympc/protocol/spdz/spdz.py#L128.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Sure, I will have a look at it.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Will do so.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Sure

kamathhrishi: Done.

tests/sympc/tensor/replicatedshare_tensor_test.py
kamathhrishi: @aanurraj Done.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Done.

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -136,3 +138,30 @@ def test_hook_property(get_clients) -> None:
 
     assert (rst.T.shares[0] == x.T).all()
     assert (rst.T.shares[1] == y.T).all()
+
+
+@pytest.mark.parametrize("parties", [2, 3, 5, 7, 11])
```
rasswanth-s, line 143: Hrishi, I think we would include a sanity check in the future which prevents initializing falcon with other than `3`  parties. This test scenario would conflict that change.

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -136,3 +138,30 @@ def test_hook_property(get_clients) -> None:
...
+
+    secret = 42.32
+
+    a = MPCTensor(secret=secret, session=session)
+
+    for share in a.share_ptrs:
+        assert len(share.get_shares().get()) == (len(parties) - 1)
+
+
+@pytest.mark.parametrize("parties", [2, 3, 5, 7, 11])
```
rasswanth-s, line 157: Same here

tests/sympc/tensor/replicatedshare_tensor_test.py
rasswanth-s, line 143: Hrishi , I think we agreed to include a sanity check in the future preventing initializing session of falcon with other `3` parties. This would conflict that scenario.

tests/sympc/tensor/replicatedshare_tensor_test.py
rasswanth-s, line 157: Same here Hrishi

tests/sympc/tensor/replicatedshare_tensor_test.py
kamathhrishi: Hmmmm, true. RIght now using Falcon as the protocol is just a proxy for testing RSTensor. I think the default protocol we discussed in https://github.com/OpenMined/SyMPC/issues/218 should serve the purpose when the protocol does not serve a specific number of parties.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Done

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +231,64 @@ def ne(self, y):
...
+        ptr_list = []
+        for i in range(0, len(parties)):
+            party_ptrs = []
+
+            for j in range(i, i + nshares):
```
rasswanth-s, line 279: MAYBE : Hrishi , I think we could incorporate the rank variable which would make it similar to `ShareTensor` distribution code.
Similar like this Pseudo code : 
<pre>
<code>
parties = session.parties
nr_parties = len(parties)
for i,party in enumerate(parties):
      rank =i
      rs_shares=[]
      for j in range(nshares):
             rs_shares.append(shares[rank].tensor)
             rank = (rank+1)%nr_parties
      tensor =ReplicatedShareTensor(shares=rs_shares,....).send(party)
</code>
</pre>

src/sympc/tensor/mpc_tensor.py
```diff
@@ -368,39 +366,22 @@ def reconstruct(
...
-            return res
-
-        request = _request_and_get
-        request_wrap = parallel_execution(request)
-
-        args = [[share] for share in self.share_ptrs]
-        local_shares = request_wrap(args)
-
-        shares = [share.tensor for share in local_shares]
+        print("MPC plaintext: ", plaintext)
```
gmuraru, line 380: MAYBE: Remove this

src/sympc/tensor/mpc_tensor.py
```diff
@@ -368,39 +366,22 @@ def reconstruct(
 
         Args:
             decode (bool): True if decode using FixedPointEncoder. Defaults to True
-            get_shares (bool): True if get shares. Defaults to False.
+            get_shares (bool): Retrieve only shares.
 
         Returns:
             torch.Tensor. The secret reconstructed.
         """
+        plaintext = self.session.protocol.share_class.reconstruct(
```
gmuraru, line 374: MAYBE: Could we rename this to something else? (since it might not be the plaintext result?)

src/sympc/tensor/mpc_tensor.py
kamathhrishi: Oops, that's a debug print.

gmuraru, line 380: MAYBE: Remove this

src/sympc/tensor/mpc_tensor.py
kamathhrishi: Sure, result would do.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -64,6 +66,8 @@ def __init__(
         self.ring_size = ring_size
 
         self.config = config
+        self.fp_encoder = None
```
gmuraru, line 69: MAYBE: Remove this

src/sympc/tensor/mpc_tensor.py
kamathhrishi: Debug print.

src/sympc/tensor/replicatedshare_tensor.py
gmuraru: I know it is not the scope of this PR - but could we also remove the `AUTOGRAD_IS_ON` from line 40 (I think that attribute is now in the MPC Level)

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +252,158 @@ def ne(self, y):
...
+
+        shares = []
+
+        if isinstance(shares2, torch.Tensor):
+
+            shares = [shares1] + shares2
+
+        else:
+
+            shares = [shares1] + shares2
```
gmuraru, line 280: Remove ```if...else``` since both paths are the same

kamathhrishi:
> Hrishi , this sanity check is done at protocol level itself during initialization.I think would not need it here.
> https://github.com/OpenMined/SyMPC/blob/16fa17ab61fc37f0aba2f565bda812ae55eb1cdc/src/sympc/protocol/falcon/falcon.py#L33

Will do so.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +252,158 @@ def ne(self, y):
...
+
+        if get_shares:
+            return shares
+
+        return sum(shares)
+
+    @staticmethod
+    def reconstruct_malicious(
```
gmuraru, line 288: Could you have only a reconstruct method and depending on the protocol type you call ```reconstruct_malicious``` or ```reconstruct_semi_honest``` (I am also thinking that this reconstruct malicious/semi-honest should live somewhere in the protocol, but let's keep it here for the moment).

Also, since ```reconstruct_semi_honest/malicious``` would be accessed only internally you can have them with a ```_```(underscore) as a prefix

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +252,158 @@ def ne(self, y):
...
+            shares1 = share_ptrs[(party_rank)].get_shares()[0].get()
+            shares2 = share_ptrs[(party_rank + 1) % (nparties)].get_shares().get()
+
+            shares = []
+
+            if isinstance(shares2, torch.Tensor):
+                shares = [shares1] + [shares2]
+
+            else:
+                shares = [shares1] + shares2
```
gmuraru, line 320: MAYBE: Remove ```if...else``` because shares2 should always be a list

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +252,158 @@ def ne(self, y):
...
+
+            if get_shares:
+                return shares
+
+            all_shares.append(shares)
+
+        # reconstruct shares from all parties and verify
+        value = None
```
gmuraru, line 328: MAYBE: 
```
ref_shares, *to_check_shares = all_shares
ref_value = sum(ref_shares)
for shares in to_check_shares:
    if ref_value != sum(shares):
       raise...
```

src/sympc/tensor/replicatedshare_tensor.py
aanurraj: hmm we can have a parameter in reconstructing for RST, for the type of reconstruction

gmuraru: **LE**: Add only the ```_``` to the method.
I did not see the ```reconstruct```

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -136,3 +141,64 @@ def test_hook_property(get_clients) -> None:
...
+    secret = 42.32
+
+    a = MPCTensor(secret=secret, session=session)
+
+    assert np.allclose(secret, a.reconstruct(), atol=1e-5)
+
+
+@pytest.mark.parametrize("parties", [2, 5, 11])
+def test_share_distribution(get_clients, parties):
+
```
gmuraru, line 178: NITPICK: Remove empty line

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -136,3 +141,64 @@ def test_hook_property(get_clients) -> None:
...
+    parties = get_clients(parties)
+    protocol = Falcon(security)
+    session = Session(protocol=protocol, parties=parties)
+    SessionManager.setup_mpc(session)
+
+    secret = 42.32
+
+    a = MPCTensor(secret=secret, session=session)
+
+    assert np.allclose(secret, a.reconstruct(), atol=1e-5)
```
gmuraru, line 173: Q: Since you did not run any operation, could you remove `atol`? (I think it might work also without it)

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -136,3 +141,64 @@ def test_hook_property(get_clients) -> None:
 
     assert (rst.T.shares[0] == x.T).all()
     assert (rst.T.shares[1] == y.T).all()
+
+
+@pytest.mark.parametrize("parties", [2, 5, 11])
+def test_distribute_sharecount(get_clients, parties) -> None:
```
gmuraru, line 147: MAYBE: This test seems similar to the `test_share_distribution`?

gmuraru (changes requested):
Some comments

tests/sympc/tensor/replicatedshare_tensor_test.py
kamathhrishi:
```
E       assert 42.32 == tensor([42.3200])
E         +42.32
E         -tensor([42.3200])
```

It throws this error.

kamathhrishi: It throws an error 

`>       assert secret==a.reconstruct().item()
E       assert 42.32 == 42.31999206542969
E         +42.32
E         -42.31999206542969`

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: I had completely refactored the code because there was some weird behaviour calling get each time in the look. The new code is concise I guess.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +249,140 @@ def ne(self, y):
...
+            share_ptrs (List[RSTensorPointers]): List of RSTensor pointers.
+            get_shares (bool): Retrieve only shares.
+
+        Returns:
+            reconstructed_value (torch.Tensor): Reconstructed value.
+        """
+        shares1 = share_ptrs[0].get_shares()[0].get()
+        shares2 = share_ptrs[1].get_shares().get()
+
+        shares = []
```
gmuraru, line 269: MAYBE: Remove this

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +249,140 @@ def ne(self, y):
...
+
+        Raises:
+            ValueError: When parties share values are not equal.
+        """
+        nparties = len(share_ptrs)
+
+        # Get shares from all parties
+        all_shares = []
+        for party_rank in range(nparties):
+            all_shares.append(share_ptrs[party_rank].get_shares().get())
```
gmuraru, line 300: MAYBE: Why not calling directly get? (without `get_shares`?)
Here you are doing 2 network calls: ```get_shares``` + `get`.

I am thinking that maybe we can remain only with `get`

gmuraru: And the call from line 318 would be changes to something like:
```return [shares in rsts.get_shares()]```

File: src/sympc/tensor/replicatedshare_tensor.py [hunk 23]
```diff
@@ -209,6 +249,140 @@ def ne(self, y):
...
+    @staticmethod
+    def reconstruct(
+        share_ptrs: List["ReplicatedSharedTensor"],
+        get_shares: bool = False,
+        security_type: str = "semi-honest",
+    ) -> Union[torch.Tensor, List[torch.Tensor]]:
+        """Reconstruct value from shares.
+
+        Args:
+            share_ptrs (List[RSTensorPointers]): List of RSTensor pointers.
```
gmuraru, line 331: Q: Is ShareTensorPointer in the ShareTensor? (I do not know if we should use here the concept of "Pointer") - if it is like that there then is ok, if not we might stick with the same naming convention that we used in ShareTensor or change ShareTensor to follow this convention

src/sympc/tensor/replicatedshare_tensor.py
gmuraru: LE: Ignore me, it might be better to do it like this since we would serialize/deserialize a tensor rather than a ReplicatedSharingTensor.

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -136,3 +140,48 @@ def test_hook_property(get_clients) -> None:
...
+
+    secret = 42.32
+
+    a = MPCTensor(secret=secret, session=session)
+
+    assert np.allclose(secret, a.reconstruct(), atol=1e-5)
+
+
+@pytest.mark.parametrize("parties", [2, 5, 11])
+def test_share_distribution(get_clients, parties):
```
gmuraru, line 161: MAYBE: The test name does not correctly show what is done in the test: ```test_share_distribution_number_shares``` might indicate better what we test?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -201,6 +220,27 @@ def eq(self, y: Any):
...
+
+    def __setitem__(self, key: int, newvalue: torch.Tensor) -> None:
+        """Allows to set share value to new value.
+
+        Args:
+            key (int): The share to be retrieved.
+            newvalue (torch.Tensor): New value of share.
+
+        """
+        self.shares = newvalue
```
rasswanth-s, line 242: I think here it should be `self.shares[key] = newvalue`.

kamathhrishi: You are right. Thanks.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -74,30 +75,48 @@ def __init__(
...
 
-    def _encode(self, data):
+    def _encode(self, data: torch.Tensor) -> torch.Tensor:
+        """Encode via FixedPointEncoder.
+
+        Args:
+            data (torch.Tensor): Tensor to be encoded
+
+        Returns:
+            encoded_data (torch.Tensor): Decoded values
```
rasswanth-s, line 85: I think here it should be `Encoded vales` in the docstring.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +249,138 @@ def ne(self, y):
...
+        get_shares: bool = False,
+    ) -> Union[torch.Tensor, List[torch.Tensor]]:
+        """Reconstruct value from shares.
+
+        Args:
+            share_ptrs (List[ReplicatedSharedTensor]): List of RSTensor pointers.
+            get_shares (bool): Retrieve only shares.
+
+        Returns:
+            reconstructed_value (torch.Tensor): Reconstructed value.
```
rasswanth-s, line 264: I think return should also include `shares :List[torch.Tensor]`

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +249,138 @@ def ne(self, y):
...
```
rasswanth-s, line 288: Same here for the return value in docstring.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +249,138 @@ def ne(self, y):
...
+    ) -> Union[torch.Tensor, List[torch.Tensor]]:
+        """Reconstruct value from shares.
+
+        Args:
+            share_ptrs (List[ReplicatedSharedTensor]): List of RSTensor pointers.
+            security_type (str): Type of security followed by protocol.
+            get_shares (bool): Retrieve only shares.
+
+        Returns:
+            reconstructed_value (torch.Tensor): Reconstructed value.
```
rasswanth-s, line 334: Same here for the `return type` in docstring.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Its more or less the same.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Hmmm, I followed the convention of MPCTensor where it was just reconstructed value, so thought of sticking to that.

kamathhrishi: But, yeah I will modify it.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +249,138 @@ def ne(self, y):
...
+        parties = session.parties
+
+        nshares = len(parties) - 1
+
+        ptr_list = []
+        for i in range(len(parties)):
+            party_ptrs = []
+
+            for j in range(i, i + nshares):
+                tensor = shares[j % (nshares + 1)].tensor
```
rasswanth-s, line 372: I think we should use here `nparties` instead `nshares`
Similar like : `tensor = shares[j % nparties].tensor` like you did for all shares
https://github.com/OpenMined/SyMPC/blob/fe974c98eaf506060f3148ed6d213c2e7be69488/src/sympc/tensor/replicatedshare_tensor.py#L304
We made an explicit assumption that 
nshares= nparties-1
I think for generalized RSS ,its bit different.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: No, I won't because it ends up making docstring too long giving lint errors. Let's keep it that way for now.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +249,138 @@ def ne(self, y):
...
+            List of ShareTensor.
+
+        """
+        parties = session.parties
+
+        nshares = len(parties) - 1
+
+        ptr_list = []
+        for i in range(len(parties)):
+            party_ptrs = []
```
rasswanth-s, line 369: MAYBE : We could have it as `party_shares` as we are only appending tensor shares.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Same as the above comment. Let's leave it there to keep docstring short.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: No, it's the same for generalized RSTensor.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Okay

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -209,6 +249,138 @@ def ne(self, y):
...
+            return ReplicatedSharedTensor.__reconstruct_malicious(
+                share_ptrs, get_shares
+            )
+
+        elif security_type == "semi-honest":
+            return ReplicatedSharedTensor.__reconstruct_semi_honest(
+                share_ptrs, get_shares
+            )
+
+        raise ValueError("Invalid security Type")
```
rasswanth-s, line 349: I think we could remove this, because during protocol instance creation itself,there is a sanity check which prevents invalid `security_type` input
https://github.com/OpenMined/SyMPC/blob/16fa17ab61fc37f0aba2f565bda812ae55eb1cdc/src/sympc/protocol/falcon/falcon.py#L33

kamathhrishi: I feel it's better because in case there is a wrong security type somewhere and somehow. It's better to throw an error than having the function return a None which will be harder to debug. 
It could help even during development, not necessarily when an end-user using SyMPC.

src/sympc/tensor/replicatedshare_tensor.py
rasswanth-s: I believe that it is protocol dependent Hrishi.Considering a scenario of 5 parties.If the share distribution is done according to our setting.`nshares`= `nparties-1`. The protocol could not guarantee malicious security for honest majority as two parties can always collude to reveal the private input.I think such protocol would do the share distribution differently to ensure honest majority. I might be wrong here too.

kamathhrishi: Okay, let's change it later I guess. Right now for Falcon and ABY3 it works just fine.

kamathhrishi: Clearly defining what this behaviour will be like will require a good amount of literature survey and we aren't sure if we will use all of them. As long as this is good enough for a default protocol, we are good.

rasswanth-s (approved):
Looks good to me.

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -12,11 +11,18 @@ def test_share_class() -> None:
 
 
 def test_session() -> None:
-    protocol = Falcon("malicious")
+    protocol = Falcon("semi-honest")
     session = Session(protocol=protocol)
     assert type(session.protocol) == Falcon
 
 
+def test_exception_malicious_lessparties(get_clients, parties=2) -> None:
```
gmuraru, line 19: NITPICK: ```less_parties```

gmuraru (approved):
LGTM!

PR #221, opened 2021-06-06
Max pooling 2D

### LINKED ISSUE #201: Add  max_pool2d support
## Feature Description
Add  max_pool2d support currently required for training.

gmuraru:
## Description
Add the `max_pooling2D` operation.
Fixes #201 

TBD:
- [x] Add the Backward step

## Affected Dependencies
Training

## How has this been tested?
- Tests should be added

## Checklist

### DIFF
--- a/src/sympc/module/nn/__init__.py
+++ b/src/sympc/module/nn/__init__.py
@@ -1,8 +1,9 @@
 """Modules and specifc functionalities needed in a Neural Network."""
 
 from .conv import Conv2d
+from .functional import max_pool2d
 from .functional import mse_loss
 from .functional import relu
 from .linear import Linear
 
-__all__ = ["relu", "mse_loss", "Linear", "Conv2d"]
+__all__ = ["relu", "mse_loss", "max_pool2d", "Linear", "Conv2d"]
--- a/src/sympc/module/nn/functional.py
+++ b/src/sympc/module/nn/functional.py
@@ -80,7 +80,7 @@ def _sanity_check_max_pool2d(
         A 4 element type with types Tuple[int, int] representing the converted parameters.
 
     Raises:
-        ValueError: if the passed in parameters are not
+        ValueError: if the parameters are not passing the sanity check
     """
     if isinstance(kernel_size, int):
         kernel_size = (kernel_size, kernel_size)
@@ -217,7 +217,8 @@ def max_pool2d(
     stride: Optional[Union[int, Tuple[int, int]]] = None,
     padding: Union[int, Tuple[int, int]] = 0,
     dilation: Union[int, Tuple[int, int]] = 1,
-) -> MPCTensor:
+    return_indices: bool = False,
+) -> Union[MPCTensor, Tuple[MPCTensor, MPCTensor]]:
     """Compute the max pool for a tensor with 2 dimension.
 
     Args:
@@ -230,13 +231,27 @@ def max_pool2d(
             in case it is passed as an integer then that specific value is used for height and width
         dilation (Union[int, Tuple[int, int]]): the dilation size
             in case it is passed as an integer then that specific value is used for height and width
+        return_indices (bool): to return the indices of the max values
 
     Returns:
         A tuple representing maximum values and the indices (as a one hot encoding
+
+    Raises:
+        ValueError: if the kernel size is bigger than the input
     """
     kernel_size, stride, padding, dilation = _sanity_check_max_pool2d(
         kernel_size, stride, padding, dilation
     )
+
+    if (
+        x.shape[-2] + 2 * padding[0] < kernel_size[0]
+        or x.shape[-1] + 2 * padding[1] < kernel_size[1]
+    ):
+        raise ValueError(
+            f"Kernel size ({kernel_size}) has more elements on an axis than "
+            f"input shape ({x.shape}) considering padding of {padding}"
+        )
+
     x_reshaped = _reshape_max_pool2d(
         x, kernel_size=kernel_size, stride=stride, padding=padding, dilation=dilation
     )
@@ -244,14 +259,17 @@ def max_pool2d(
     res_max_columns, columns = x_reshaped.max(dim=-1, one_hot=True)
     res_max, rows = res_max_columns.max(dim=-1, one_hot=True)
 
-    indices = columns * rows.unsqueeze(-1)
-
     output_shape = x.shape[:-2] + (
         (x.shape[-2] - kernel_size[0] + 2 * padding[0]) // stride[0] + 1,
         (x.shape[-1] - kernel_size[1] + 2 * padding[1]) // stride[1] + 1,
     )
 
-    return res_max.reshape(*output_shape), indices.reshape(output_shape + kernel_size)
+    res = res_max.reshape(*output_shape)
+    if return_indices:
+        indices = columns * rows.unsqueeze(-1)
+        res = (res, indices.reshape(output_shape + kernel_size))
+
+    return res
 
 
 def max_pool2d_backward_helper(
@@ -260,7 +278,7 @@ def max_pool2d_backward_helper(
     kernel_size: Tuple[int, int],
     stride: Tuple[int, int],
     padding: Tuple[int, int],
-):
+) -> ShareTensor:
     """Helper function to compute the gradient needed to be passed to the parent node.
 
     Args:
@@ -272,11 +290,21 @@ def max_pool2d_backward_helper(
 
     Returns:
         A ShareTensor specific for the computed gradient
+
+    Raises:
+        ValueError: if the input shape (taken into consideration the padding) is smaller than the
+            kernel shape
     """
     session = get_session(str(grads_share.session_uuid))
 
     res_shape = input_shape[:-2]
     res_shape += (input_shape[-2] + 2 * padding[0], input_shape[-1] + 2 * padding[1])
+
+    if res_shape[-2] < kernel_size[0] or res_shape[-1] < kernel_size[1]:
+        raise ValueError(
+            f"Kernel size ({kernel_size}) has more elements on an axis than "
+            f"input shape ({res_shape}) considering padding of {padding}"
+        )
 
     tensor_type = session.tensor_type
     tensor = torch.zeros(res_shape, dtype=tensor_type)
--- a/src/sympc/module/nn/functional.py
+++ b/src/sympc/module/nn/functional.py
@@ -181,9 +181,9 @@ def helper_max_pool2d_reshape(
 
     padding_value = 0
     if session.rank == 0:
-        # ATTENTION: Min value for max_pool2d that works -2**8
-        # This is because 2 is the base and 16 is the precision 2 ** 24 / (2 ** 16) = 2 ** 9
-        padding_value = -(2 ** 24)
+        # ATTENTION: Min value for max_pool2d that works -2**6
+        # This is because 2 is the base and 16 is the precision 2 ** 22 / (2 ** 16) = 2 ** 6
+        padding_value = -(2 ** 22)
 
     tensor = np.pad(tensor, padding, mode="constant", constant_values=padding_value)
 
--- a/src/sympc/module/nn/functional.py
+++ b/src/sympc/module/nn/functional.py
@@ -218,7 +218,7 @@ def max_pool2d(
     padding: Union[int, Tuple[int, int]] = 0,
     dilation: Union[int, Tuple[int, int]] = 1,
 ) -> MPCTensor:
-    """Compute the max poold for a tensor with 2 dimension.
+    """Compute the max pool for a tensor with 2 dimension.
 
     Args:
         x (MPCTensor): the MPCTensor on which to apply the operation
--- a/src/sympc/module/nn/functional.py
+++ b/src/sympc/module/nn/functional.py
@@ -181,9 +181,9 @@ def helper_max_pool2d_reshape(
 
     padding_value = 0
     if session.rank == 0:
-        # ATTENTION: Min value for max_pool2d that works -2**6
-        # This is because 2 is the base and 16 is the precision 2 ** 22 / (2 ** 16) = 2 ** 6
-        padding_value = -(2 ** 22)
+        # ATTENTION: Min value for max_pool2d that works -2**8
+        # This is because 2 is the base and 16 is the precision 2 ** 24 / (2 ** 16) = 2 ** 8
+        padding_value = -(2 ** 24)
 
     tensor = np.pad(tensor, padding, mode="constant", constant_values=padding_value)
 
--- a/src/sympc/module/nn/functional.py
+++ b/src/sympc/module/nn/functional.py
@@ -181,8 +181,8 @@ def helper_max_pool2d_reshape(
 
     padding_value = 0
     if session.rank == 0:
-        # ATTENTION: Min value for max_pool2d that works -50
-        padding_value = -50
+        # ATTENTION: Min value for max_pool2d that works -25
+        padding_value = -25
 
     tensor = np.pad(tensor, padding, mode="constant", constant_values=padding_value)
 
--- a/src/sympc/module/nn/functional.py
+++ b/src/sympc/module/nn/functional.py
@@ -51,6 +51,72 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
     return result
 
 
+Kernel2D = Tuple[int, int]
+Stride2D = Tuple[int, int]
+Padding2D = Tuple[int, int]
+Dilation2D = Tuple[int, int]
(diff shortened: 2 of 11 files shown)


src/sympc/tensor/grads/grad_functions.py
```diff
@@ -511,39 +482,28 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> Tuple[MPCTensor]:
             grad (MPCTensor): The gradient that came from the child nodes
 
         Returns:
-            (input_grad, weight_grad) (Tuple[MPCTensor]): The gradients passed
+            (input_grad, weight_grad) (Tuple[MPCTensor, MPCTensor]): The gradients passed
             to the input and kernal nodes.
         """
-        input = ctx["input"]
```
gmuraru, line 517: Changed this to "x" because `input` is a Python internal reserved word

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,142 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+
+
+def max_pool2d(
+    x: MPCTensor,
+    kernel_size: Union[int, Tuple[int, int]],
+    stride: Optional[Union[int, Tuple[int, int]]] = None,
+    padding: Union[int, Tuple[int, int]] = 0,
+    dilation: Union[int, Tuple[int, int]] = 1,
+) -> MPCTensor:
+    """Compute the max poold for a tensor with 2 dimension.
```
aanurraj, line 148: TYPO: pool

aanurraj (commented):
Its looks good, may be we need tests for helper functions too :D

src/sympc/tensor/grads/grad_functions.py
```diff
@@ -447,53 +447,24 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> MPCTensor:
...
+        dilation: Union[int, Tuple[int, int]] = 1,
+        groups: int = 1,
     ) -> MPCTensor:
         """Perform the feedforward and compute the result for the conv2d operation.
 
         Args:
             ctx (Dict[str, Any]): Context used to save information needed in the backward pass
-            input: the input
+            x: the input
             weight: the convolution kernel
```
gmuraru, line 466: This also need adding type hints but I will open an issue for them

src/sympc/tensor/grads/grad_functions.py
```diff
@@ -502,14 +473,14 @@ def forward(
         Returns:
             (MPCTensor): The result of the conv2d operation
         """
-        ctx["input"] = input
```
gmuraru, line 505: `input` is a python specific keyword and this is not a good idea to use it

src/sympc/tensor/grads/grad_functions.py
```diff
@@ -520,39 +491,28 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> Tuple[MPCTensor]:
...
-        )(args)
-        grad_input_padding = MPCTensor(shares=shares, session=grad.session)
-
-        output_padding_tensor = grad_input_padding.reconstruct()
-        output_padding_tensor /= grad.session.nr_parties
-        output_padding = tuple(output_padding_tensor.to(torch.int).tolist())
+        min_batch = x.shape[0]
+
+        output_padding = torch.nn.grad._grad_input_padding(
+            grad_output=torch.empty(grad.shape),
```
gmuraru, line 509: `grad_output` param is only used to get the ```shape/dim``` there is no need

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,323 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+    if len(kernel_size) != 2:
+        raise ValueError("Kernel_size should have only 2 dimensions")
+
+    if stride is None:
+        stride = kernel_size
+
+    if isinstance(stride, int):
+        stride = (stride, stride)
+
+    if len(stride) != 2:
```
kamathhrishi, line 329: Maybe: Sanity checks as a separate function?

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,323 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+        dilation (Union[int, Tuple[int, int]]): the dilation size
+            in case it is passed as an integer then that specific value is used for height and width
+
+    Returns:
+        A tuple representing maximum values and the indices (as a one hot encoding
+
+    Raises:
+        ValueError: in case kernel_size is passed as a tuple with a number of elements != 2
+    """
+    if isinstance(kernel_size, int):
```
kamathhrishi, line 174: Maybe: A separate function for sanity checks? Lines 174 to 205. If backward and forward have the same sanity checks, you can combine them into one function and have the rest of them in the function. Or have two separate sanity check functions for backward and forward depending on the overlap. Just to keep the code clean and easier to understand.

Syzygianinfern0: Yes, I agree

src/sympc/tensor/grads/grad_functions.py
aanurraj: hmm didn't understand this.

src/sympc/tensor/grads/grad_functions.py
```diff
@@ -667,6 +625,86 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> MPCTensor:
         return res_grad
 
 
+class GradMaxPool2D(GradFunc):
+    """The Reshape gradient function."""
```
aanurraj, line 629:
```suggestion
    """The MaxPool2D gradient function."""
```

tests/sympc/module/nn/functional_test.py
```diff
@@ -41,3 +44,39 @@ def test_mse_loss(get_clients, reduction) -> None:
...
+    (3, 2, 0),
+    (3, 2, 1),
+    (3, 3, 0),
+    (3, 3, 1),
+]
+
+
+@pytest.mark.parametrize("kernel_size, stride, padding", POSSIBLE_CONFIGS_MAXPOOL_2D)
+def test_max_pool2d(get_clients, kernel_size, stride, padding) -> None:
+    clients = get_clients(2)
```
aanurraj, line 66: are we using fss_comp somewhere?

tests/sympc/module/nn/functional_test.py
```diff
@@ -41,3 +44,39 @@ def test_mse_loss(get_clients, reduction) -> None:
...
+    mpc = MPCTensor(secret=secret, session=session)
+
+    res, _ = sympc.module.nn.max_pool2d(
+        mpc, kernel_size=kernel_size, stride=stride, padding=padding
+    )
+    res_expected = torch.max_pool2d(
+        secret, kernel_size=kernel_size, stride=stride, padding=padding
+    )
+
+    assert np.allclose(res.reconstruct(), res_expected, atol=1e-4)
```
aanurraj, line 82: 👍🏻

tests/sympc/module/nn/functional_test.py
gmuraru: Yes. ```max_pool2d``` needs comparison to check what is the bigger element

gmuraru:
@aanurraj for the `torch.empty` gradient - you require for padding only the shape and we already know that (if you look in the ```_input_padding``` function from torch you will see that they only call `shape/dim` methods on the gradient input.

src/sympc/tensor/grads/grad_functions.py
gmuraru: grad_output parameter is only used for the shape so there is no need to really compute it

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,327 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+    Returns:
+        The gradient that should be backpropagated (MPCTensor)
+
+    Raises:
+        ValueError: In case some of the values for the parameters are not supported
+    """
+    kernel_size, stride, padding, dilation = _sanity_check_max_pool2d(
+        kernel_size, stride, padding, dilation
+    )
+    if len(grad.shape) != 4 and len(grad.shape) != 3:
```
aanurraj, line 354: Shouldn't it be or? instead of and?

aanurraj (commented):
Left comments and this looks great thanks :D

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,327 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+    if isinstance(stride, int):
+        stride = (stride, stride)
+
+    if len(stride) != 2:
+        raise ValueError("Stride should have only 2 dimensions")
+
+    if isinstance(padding, int):
+        padding = (padding, padding)
+
+    if padding[0] > kernel_size[0] or padding[1] > kernel_size[1]:
```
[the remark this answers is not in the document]
gmuraru: Nope, it is an error.

gmuraru:
> Left comments and this looks great thanks :D

Thank you!

src/sympc/module/nn/functional.py
gmuraru: What we want to check here if `grad` shape is 3 or 4 -- the if will be triggered only if it is neither 3 or 4.
If we apply De Morgan rule (with double negation) we could have an or - but I think it is ok to have it with an `and`.

Or do you think it is not understandable what was the intention here?

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,327 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+            in case it is passed as an integer then that specific value is used for height and width
+        padding (Union[int, Tuple[int, int]]): the padding size
+            in case it is passed as an integer then that specific value is used for height and width
+        dilation (Union[int, Tuple[int, int]]): the dilation size
+            in case it is passed as an integer then that specific value is used for height and width
+
+    Returns:
+        A tuple representing maximum values and the indices (as a one hot encoding
+    """
+    kernel_size, stride, padding, dilation = _sanity_check_max_pool2d(
```
kamathhrishi, line 238: I think we need a check to see if the input x is bigger than kernel size. Sometimes this error does occur unintentionally when designing networks. Also, add a test for it. 

Or did I miss this check somewhere.

kamathhrishi: The way PyTorch handles it. 
https://discuss.pytorch.org/t/kernel-size-cant-greater-than-actual-input-size/21579

tests/sympc/tensor/grad/grad_functions_test.py
```diff
@@ -641,3 +614,109 @@ def test_forward(get_clients) -> None:
...
+    x = secret.share(parties=parties)
+
+    with pytest.raises(ValueError):
+        GradMaxPool2D.forward({}, x, kernel_size=2, dilation=2)
+
+    with pytest.raises(ValueError):
+        GradMaxPool2D.forward({}, x, kernel_size=2, dilation=(1, 2))
+
+
+POSSIBLE_CONFIGS_MAXPOOL_2D = [
```
kamathhrishi, line 635: Maybe: Add a test with rectangular kernel size? A rectangular kernel is used sometimes. Like for example, speech recognition. 
In case it's not supported for some reason we can throw an error. But, I noticed that the Conv2d was accepting tuple or int as kernel size.

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,327 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+    return res_max.reshape(*output_shape), indices.reshape(output_shape + kernel_size)
+
+
+def max_pool2d_backward_helper(
+    input_shape: Tuple[int],
+    grads_share: ShareTensor,
+    kernel_size: Tuple[int, int],
+    stride: Tuple[int, int],
+    padding: Tuple[int, int],
+):
```
kamathhrishi, line 264: Missing return typehint.

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,327 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+    )
+
+    window_view_share = window_view_share.reshape(-1, *kernel_size)
+
+    res_share = ShareTensor(config=x.config)
+    res_share.tensor = window_view_share
+    return res_share
+
+
+def max_pool2d(
```
kamathhrishi, line 215: Also add a check somewhere to ensure MPCTensor with ShareTensor as share_class is being passed, since we do not yet support RSTensor with conv2d. You can raise a NotImplementedError.

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,327 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+    padding = [(0, 0)] * len(tensor.shape[:-2]) + [
+        (padding[0], padding[0]),
+        (padding[1], padding[1]),
+    ]
+    tensor_type = session.tensor_type
+
+    padding_value = 0
+    if session.rank == 0:
+        # ATTENTION: Min value for max_pool2d that works -25
+        padding_value = -25
```
kamathhrishi, line 185: Curious, what does this do?

src/sympc/module/nn/__init__.py
```diff
@@ -1,8 +1,9 @@
 """Modules and specifc functionalities needed in a Neural Network."""
 
 from .conv import Conv2d
+from .functional import max_pool2d
 from .functional import mse_loss
 from .functional import relu
 from .linear import Linear
 
-__all__ = ["relu", "mse_loss", "Linear", "Conv2d"]
+__all__ = ["relu", "mse_loss", "max_pool2d", "Linear", "Conv2d"]
```
Syzygianinfern0, line 9: Why do we not have an `nn.MaxPool2d` that calls `F.max_pool2d`? It would feel more intuitive to translate PyTorch models to SyMPC models.

[PyTorch src](https://pytorch.org/docs/stable/_modules/torch/nn/modules/pooling.html#:~:text=class%20MaxPool2d(_MaxPoolNd)%3A)

tests/sympc/module/nn/functional_test.py
Syzygianinfern0: 1e-4 and tests pass! 🤯
That's great!

Syzygianinfern0 (approved):
Great! 🙌

src/sympc/module/nn/functional.py
```diff
@@ -37,3 +49,327 @@ def mse_loss(pred: MPCTensor, target: MPCTensor, reduction: str = "mean") -> MPC
...
+    res_share.tensor = window_view_share
+    return res_share
+
+
+def max_pool2d(
+    x: MPCTensor,
+    kernel_size: Union[int, Tuple[int, int]],
+    stride: Optional[Union[int, Tuple[int, int]]] = None,
+    padding: Union[int, Tuple[int, int]] = 0,
+    dilation: Union[int, Tuple[int, int]] = 1,
```
kamathhrishi, line 220: Maybe: add argument return indices similar to Pytorch API: https://pytorch.org/docs/stable/generated/torch.nn.MaxPool2d.html

src/sympc/module/nn/__init__.py
gmuraru: We followed the approach with ```nn.functional.max_pool2d``` (pytorch let's you use those functions from more places).

src/sympc/module/nn/functional.py
gmuraru: Select a small value for the padding.
Or could you be more specific?

src/sympc/module/nn/functional.py
gmuraru: I added - only that we can not test it because that parameter should be moved to module (like torch has MaxPool2d). For the moment, we use directly a function -- this one from here (similar to `torch.max_pool2d` - but that one does not have `return_indices` as param).

We also need to implement ```sympc.nn.moduleMaxPool2d``` similar to `torch.nn.module.MaxPool2d` (and this has the parameter `return_indices`)

src/sympc/module/nn/functional.py
gmuraru: We should do that after we have RSTensor in MPCTensor I think. And we should do that using something from python (a dispatcher class)

kamathhrishi: Cool

kamathhrishi (approved):
Looks good to me.

aanurraj (approved):

PR #226, opened 2021-06-08
Custom value for pytest workers

rasswanth-s:
## Description
Currently our CI takes a lot to time execute , setting a custom number of  workers of pytest instead of `pytest -n auto` greatly reduced the CI time.

## Affected Dependencies

## How has this been tested?
CI time duration.

## Checklist

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -13,8 +13,9 @@ notebook==6.2.0
 papermill==2.3.1
 pep8-naming==0.11.1
 pre-commit==2.11.1
-pytest==6.2.2
+pytest==6.2.4
 pytest-benchmark==3.4.1
 pytest-cov==2.10.1
+pytest-order==1.0.0
 pytest-xdist[psutil]
 safety
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -82,7 +82,7 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest tests/ --deselect=tests/notebooks -n auto
+          pytest tests/ --deselect=tests/notebooks -n 8
 
       - name: Run codecov
         if: ${{ matrix.os == 'ubuntu-latest' }}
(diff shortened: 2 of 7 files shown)


arturomf94 (approved):
Looks good. Just out of curiosity: Why is this so? Why is auto not working? And why was 8 chosen?

Syzygianinfern0:
@gmuraru can we do this: [Rerun Failed Tests](https://docs.pytest.org/en/latest/how-to/cache.html#:~:text=you%20have%20run%20only%20the%20two%20failing%20tests%20from%20the%20last%20run%2C%20while%20the%2048%20passing%20tests%20have%20not%20been%20run). Flaky tests are annoying :grimacing:.

rasswanth-s:
Yeah @Syzygianinfern0  as you pointed ,I think there are quite few flaky tests in our repository.I re-ran several times the CI today.

rasswanth-s:
@arturomf94  I observed that in our previous CI , auto would set at max `2` or `3` workers. I empirically tested the values in my repo by changing the values.I found that `8` was optimal.I am currently experimenting with the values.It actually quite worked in our repo reducing CI time from 45-50 min to around 20 minutes.I think `auto` does not quite set the max number of process.I am also doing additional reading on it.

gmuraru:
The tests are failing because it could not find a primitive in our CryptoStore (it needs some digging) - probably we would also need to do a cleanup for that - like we clear the node store, we need to also clear that crypto primitive store.
node.session.crypto_store (if it is not None) and call clear on it -- although I am not 100% sure this would solve the problem

requirements.dev.txt
```diff
@@ -13,8 +13,9 @@ notebook==6.2.0
 papermill==2.3.1
 pep8-naming==0.11.1
 pre-commit==2.11.1
-pytest==6.2.2
+pytest==6.2.4
 pytest-benchmark==3.4.1
 pytest-cov==2.10.1
+pytest-order==1.0.0
```
gmuraru, line 19: :heart:

gmuraru (approved):
LGTM!

PR #228, opened 2021-06-09
Implementation of PRRS(Pseudo-Random Random Share)

### LINKED ISSUE #227: Correlated Randomness : RandomShare
## Feature Description
Implement a functionality such that parties could agree on a random value without knowing the actual value locally without interaction.

## Is your feature request related to a problem?
Falcon require 2-out-of-3 randomness.
We could also port it for ShareTensor.
## Additional Context
[FALCON paper](https://arxiv.org/pdf/2004.02229.pdf)
It requires one initial setup of communicating shared seeds,then all the share generation could be done locally.

rasswanth-s:
## Description
PRRS (Pseudo-Random Random Share) is implemented such that parties could compute share values of random number.fixes #227 .It is an n-out-of-n sharing for ShareTensor and 2-out-3 sharing for RSTensor.

## Affected Dependencies

## Checklist


src/sympc/session/session.py
```diff
@@ -204,12 +205,64 @@ def przs_generate_random_share(
...
-            session_uuid=self.uuid,
-            config=Config(encoder_precision=0),
+        if self.protocol.share_class == ShareTensor:
+            # It has encoder_precision = 0 such that the value would not be encoded
+            share = ShareTensor(
+                data=current_share - next_share,
+                session_uuid=self.uuid,
+                config=Config(encoder_precision=0),
+            )
+        else:
```
aanurraj, line 215: add an el if condition here and may be, raise an error if share_class is not set.

src/sympc/session/session.py
```diff
@@ -204,12 +205,64 @@ def przs_generate_random_share(
...
+        )  # We also invoke second generator as it operates in CTR mode.
+
+        if self.protocol.share_class == ShareTensor:
+            # It has encoder_precision = 0 such that the value would not be encoded
+            share = ShareTensor(
+                data=share1,
+                session_uuid=self.uuid,
+                config=Config(encoder_precision=0),
+            )
+        else:
```
aanurraj, line 260: the same condition here.

tests/sympc/session/session_test.py
```diff
@@ -66,6 +68,20 @@ def test_przs_generate_random_share(get_clients) -> None:
...
+    """Test przs_generate_random_share method from Session for ReplicatedSharedTensor."""
+    falcon = Falcon(security_type="malicious")
+    session = Session(protocol=falcon)
+    SessionManager.setup_mpc(session)
+    gen1 = get_new_generator(42)
+    gen2 = get_new_generator(43)
+    session.przs_generators = [gen1, gen2]
+    share = session.przs_generate_random_share(shape=(2, 1))
+    assert isinstance(share, ReplicatedSharedTensor)
+    target_tensor = torch.tensor(([-1540733531777602634], [2813554787685566880]))
```
aanurraj, line 81: Q: Do we need to hard code this value?

aanurraj (changes requested):

tests/sympc/session/session_test.py
rasswanth-s: It was in the previous test for PRZS , I will modify it.:thumbsup:

src/sympc/session/session.py
rasswanth-s: Yeah ,that is in the protocol metaclass , which does not allow for protocol class creation without share_class attribute.https://github.com/OpenMined/SyMPC/blob/16fa17ab61fc37f0aba2f565bda812ae55eb1cdc/src/sympc/protocol/protocol.py#L36

src/sympc/api.py
```diff
@@ -197,7 +200,15 @@
     ("sympc.protocol.spdz.spdz.div_wraps", "sympc.tensor.ShareTensor"),
     (
         "sympc.session.Session.przs_generate_random_share",
-        "sympc.tensor.ShareTensor",
+        sy.lib.misc.union.UnionGenerator[
```
gmuraru, line 203: Q: There is no other solution than using the `UnionGenerator`? (there is not something like `Union`?)

gmuraru (approved):
LGTM!

src/sympc/session/session.py
```diff
@@ -185,41 +185,98 @@ def get_protocol(self) -> Protocol:
...
             shape (Union[tuple, torch.Size]): Shape for the share.
 
         Returns:
-            Any: ShareTensor
-
+            List[torch.Tensor] : shares generated by the generators.
         """
-        from sympc.tensor import ShareTensor
-
         gen0, gen1 = self.przs_generators
```
Syzygianinfern0, line 200: Maybe: Could have these as `gen1`, `gen2` to be consistent with the rest of the changes in the PR

src/sympc/session/session.py
```diff
@@ -185,41 +185,98 @@ def get_protocol(self) -> Protocol:
...
             shape=shape,
         )
 
-        next_share = generate_random_element(
+        tensor_share2 = generate_random_element(
             tensor_type=self.tensor_type,
             generator=gen1,
             shape=shape,
         )
+        # We also invoke second generator for prrs as it operates in CTR mode.
```
Syzygianinfern0, line 213: What is CTR?

src/sympc/session/session.py
```diff
@@ -185,41 +185,98 @@ def get_protocol(self) -> Protocol:
...
-        share = ShareTensor(
-            data=current_share - next_share,
-            session_uuid=self.uuid,
-            config=Config(encoder_precision=0),
-        )
+    def przs_generate_random_share(
+        self,
+        shape: Union[tuple, torch.Size],
+    ) -> Any:
+        """Generate a random zero share using the two generators hold by a party.
```
Syzygianinfern0, line 220:
```suggestion
        """Generate a random zero share using the two generators held by a party.
```

src/sympc/session/session.py
```diff
@@ -185,41 +185,98 @@ def get_protocol(self) -> Protocol:
...
+                shares=[current_share - next_share],
+                session_uuid=self.uuid,
+                config=Config(encoder_precision=0),
+            )
+        return share
+
+    def prrs_generate_random_share(
+        self,
+        shape: Union[tuple, torch.Size],
+    ) -> Any:
```
Syzygianinfern0, line 252: Why do we not have return type as `Union[ReplicatedSharedTensor, ShareTensor]`?

src/sympc/session/session.py
```diff
@@ -185,41 +185,98 @@ def get_protocol(self) -> Protocol:
...
+        """Generate a random share using the  generators hold by a party.
```
Syzygianinfern0, line 253:
```suggestion
        """Generate a random share using the generators held by a party.
```

src/sympc/session/session.py
```diff
@@ -185,41 +185,98 @@ def get_protocol(self) -> Protocol:
...
+        """
+        from sympc.tensor import ReplicatedSharedTensor
+        from sympc.tensor import ShareTensor
+
+        current_share, next_share = self._generate_random_share(shape)
+
+        if self.protocol.share_class == ShareTensor:
+            # It has encoder_precision = 0 such that the value would not be encoded
+            share = ShareTensor(
+                data=current_share - next_share,
```
Syzygianinfern0, line 237: 👍

Syzygianinfern0 (approved):
Great work!

src/sympc/session/session.py
rasswanth-s: Thats a good suggestion.Sure , I would change it :+1:

src/sympc/session/session.py
rasswanth-s: I missed it,Sure ,I would modify it :+1:

rasswanth-s: It was in the previous PRZS code, as we import module inside the function so we  have used `Any` .It was placed inside functions to avoid circular imports.I tried solving circular imports ,but had to make a lot of changes ,thought to defer it for a later PR.

src/sympc/session/session.py
rasswanth-s: Sure, CTR means counter mode,given same   initial seeds to two generators,CTR enables us to generates same Pseudo Random number for subsequent invocations.This would enable parties given the same initial seeds to generate the same random number.

Syzygianinfern0: > CTR enables us to generates same Pseudo Random number for subsequent invocations.This would enable parties given the same initial seeds to generate the same random number.

Could we have this in the docstring of the `PR(Z/R)S` function?

Syzygianinfern0 (approved):
LGTM

kamathhrishi (approved):

aanurraj (approved):

PR #231, opened 2021-06-10
Initialization of  Default Protocol class

rasswanth-s:
## Description
Initialization of the Default Protocol class.The default protocol will contain all the primtive operations.In a session if a given protocol does not implement a specific operation.It would be reverted to the Default Protocol class.

## Affected Dependencies
Since Falcon is 3PC,Adding DefaultProtocol would make RSTensor framework agnostic

## How has this been tested?
Added tests for the same.

## Checklist

### DIFF
--- a/src/sympc/protocol/default/default.py
+++ b/src/sympc/protocol/default/default.py
@@ -1,0 +1,62 @@
+"""Default Protocol."""
+
+# stdlib
+from typing import Any
+from typing import Dict
+from typing import List
+
+from sympc.protocol.protocol import Protocol
+from sympc.tensor import ReplicatedSharedTensor
+from sympc.tensor.tensor import SyMPCTensor
+
+
+class DefaultProtocol(metaclass=Protocol):
+    """Default Protocol Implementation."""
+
+    """Used for Share Level static operations like distributing the shares."""
+    share_class: SyMPCTensor = ReplicatedSharedTensor
+    security_levels: List[str] = ["semi-honest", "malicious"]
+
+    def __init__(self, security_type: str = "semi-honest"):
+        """Initialization of the Protocol.
+
+        Args:
+            security_type : specifies the security level of the Protocol.
+
+        Raises:
+            ValueError : If invalid security_type is provided.
+        """
+        if security_type not in self.security_levels:
+            raise ValueError(f"{security_type} is not a valid security type")
+
+        self.security_type = security_type
+
+    @staticmethod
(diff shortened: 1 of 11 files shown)


src/sympc/protocol/default/default.py
```diff
@@ -0,0 +1,62 @@
...
+from sympc.protocol.protocol import Protocol
+from sympc.tensor import ReplicatedSharedTensor
+from sympc.tensor.tensor import SyMPCTensor
+
+
+class DefaultProtocol(metaclass=Protocol):
+    """Default Protocol Implementation."""
+
+    """Used for Share Level static operations like distributing the shares."""
+    share_class: SyMPCTensor = ReplicatedSharedTensor
```
gmuraru, line 17: Q: Could we have it `ShareTensor` (because all the operations that work are using `ShareTensor`)

src/sympc/protocol/default/default.py
```diff
@@ -0,0 +1,62 @@
...
+        Raises:
+            ValueError : If invalid security_type is provided.
+        """
+        if security_type not in self.security_levels:
+            raise ValueError(f"{security_type} is not a valid security type")
+
+        self.security_type = security_type
+
+    @staticmethod
+    def distribute_shares(*args: List[Any], **kwargs: Dict[str, Any]) -> Any:
```
gmuraru, line 35: Ideally, after we merge this PR, we could create another where we put all the `add`, `sub`, `mult`, `matmul`, `eq` .... operation here

gmuraru (approved):
LGTM!  Let's wait for another team member to review this before merging

src/sympc/protocol/default/default.py
gmuraru: And also the `security_levels` to be only `["semi-honest"]`

rasswanth-s: Yeah ,george  I wanted to initialize `default protocol` ,such that I could revert `RSTensor` operation when we have arbitrary number of parties,as `Falcon` is highly customized for 3PC.Shall we have it as `RSTensor` for now?.In the long run when we are implementing ops we could have both `RSTensor` and `ShareTensor`.

kamathhrishi (approved):

kamathhrishi (commented):
Will be revisiting this PR after a while.

kamathhrishi:
We can close and come back to the PR later?

Issue #158, opened 2021-05-09
Replicated Sharing Tensor (RST): Mul Operation 


kamathhrishi:
## Feature Description
Implement operation to RSS Tensor for the multiplication of private and public value.
RSS is described in #145
This requires #172 
Task list:
- [x] Semihonest
- [x] Malicious

Issue #146, opened 2021-05-09
Falcon Primitive Operations 


kamathhrishi:
## Feature Description
Primitive operations of Falcon Protocol:

- [x] Addition with public value 
- [x] Subtraction with public value 
- [x] Addition with another secret 
- [x] Subtraction with public value
- [x] Multiplication with public value 
- [x] Multiplication with another secret 

## Additional Context
[Falcon Paper](https://arxiv.org/abs/2004.02229)
This PR has a dependency on #145 and should be built on Replicated Secret Share Tensor.

gmuraru:
I think this could also be split into multiple issues?

PR #243, opened 2021-06-14
Falcon Semi-honest Integer multiplication operation 

kamathhrishi:
This PR initializes Falcon multiplication for only integers with semi-honest security. After truncation is merged, it can work on float values too. This PR also refactored RSTensor and MPCTensor to accommodate the growing codebase.

The PR added the multiplication functionality for private and public values. 

Most importantly added NotImplementedErrors, which kept throwing None and getting me confused. 

I dint' use the convention of redirecting operation to apply_function because I felt it adds too much code for a single function and wasn't necessary. 

## How has this been tested?
- Added tests 

## Checklist

Addresses issue #158 and #146


src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +65,77 @@ def __eq__(self, other: Any) -> bool:
...
+        else:
+            raise NotImplementedError(
+                "Multiplication has not been implemented for "
+                + session.protocol.security_type
+                + " setting"
+            )
+
+        vals = [result[0].get(), result[1].get(), result[2].get()]
+        # Add random mask to vals
+        shares = MPCTensor.generate_shares(
```
rasswanth-s, line 112: Q: Why are we not using PRZS to 3-out-3 sharing?

kamathhrishi: Dint see the difference. PRZS is effective only for party-party communication. Here PRZS will be essentially the same as generating shares.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +65,77 @@ def __eq__(self, other: Any) -> bool:
             return False
 
         return True
+
+    @staticmethod
+    def mul(x, y, session) -> ReplicatedSharedTensor:
+        """Master method for multiplication.
+
+        Performs Falcon's mult implementation, gets and reshares mul results and distributes shares.
```
aanurraj, line 73:
```suggestion
        Performs Falcon's mul implementation, gets and reshares mul results and distributes shares.
```

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +65,77 @@ def __eq__(self, other: Any) -> bool:
...
+
+        Returns:
+            shares (ReplicatedSharedTensor): Shares in terms of ReplicatedSharedTensor.
+
+        Raises:
+            ValueError: Raised when number of parties are not three.
+            NotImplementedError: Raised when implementation not present
+
+        """
+        if not len(session.parties) == 3:
```
aanurraj, line 90: probably != 3 😄

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +65,77 @@ def __eq__(self, other: Any) -> bool:
...
+        vals = [result[0].get(), result[1].get(), result[2].get()]
+        # Add random mask to vals
+        shares = MPCTensor.generate_shares(
+            sum(vals), 3, config=Config(encoder_base=1, encoder_precision=0)
+        )
+        shares = ReplicatedSharedTensor.distribute_shares(shares, x.session)
+        return shares
+
+    @staticmethod
+    def mul_implementation(x, y) -> List[ReplicatedSharedTensor]:
```
aanurraj, line 119: can we have a different name for this function this sounds little bit off

aanurraj, line 119: also type for x and y are missing

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -306,63 +306,97 @@ def rsub(
...
         Args:
             y: self>y
 
+        Raises:
+            NotImplementedError: Raised when implementation not present
         """
+        raise NotImplementedError
 
-    def eq(self, y: Any):
+    def eq(self, y: Any) -> bool:
```
aanurraj, line 399: Q: Are we supposed to reveal that if the value under the RSTensors is equal to y without calling a reconstruction? 
that would be a security lapse right?

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -26,3 +30,91 @@ def test_exception_malicious_less_parties(get_clients, parties=2) -> None:
...
+
+    secret1 = torch.tensor([[-100, 20, 30], [-90, 1000, 1], [1032, -323, 15]])
+
+    secret2 = 8
+
+    tensor1 = MPCTensor(secret=secret1, session=session)
+    tensor2 = MPCTensor(secret=secret2, session=session)
+
+    with pytest.raises(ValueError):
+
```
aanurraj, line 98: can you remove this empty line?

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -26,3 +30,91 @@ def test_exception_malicious_less_parties(get_clients, parties=2) -> None:
...
+    SessionManager.setup_mpc(session)
+
+    secret1 = torch.tensor([[-100, 20, 30], [-90, 1000, 1], [1032, -323, 15]])
+    secret2 = 8
+
+    tensor1 = MPCTensor(secret=secret1, session=session)
+    tensor2 = MPCTensor(secret=secret2, session=session)
+
+    with pytest.raises(NotImplementedError):
+
```
aanurraj, line 119: here also the blank line

aanurraj (commented):
check comments

aanurraj (changes requested):
check comments

src/sympc/protocol/falcon/falcon.py
rasswanth-s: True , since we rely on the orchestrator model,it does not make a difference.From a security point, falcon does not depend on third party.If we could generate local from the parties, it would be consistent with Falcon Hrishi.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: We haven't implemented this method so, can't say what lies in the future or if it will leak anything.

src/sympc/protocol/falcon/falcon.py
kamathhrishi: multiplication_protocol?

src/sympc/protocol/falcon/falcon.py
kamathhrishi: looks better

src/sympc/protocol/falcon/falcon.py
kamathhrishi: Yes, but if the orchestrator has access to all the shares. The orchestrator can anyway sum it up?

rasswanth-s: True , I totally agree,but that assumption is due to the current sympc orchestrator model. It changes  the building blocks of Falcon from security standpoint Hrishi,so I think generating from PRZS would be a better option.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: I will remove that type annotation. Its for future reference.

kamathhrishi: Oh, I am sorry for the confusion. I thought it was not implemented.

aanurraj: hm yes it should be i guess RSTensor since we don't want to reveal any bool until we call reconstruct

kamathhrishi: No, the value can't be got until get() is called. Very similar to how https://github.com/OpenMined/SyMPC/blob/2005f3393c03296bbdf18cecafc9c0582783e84f/src/sympc/tensor/share_tensor.py#L386 has it.

src/sympc/tensor/mpc_tensor.py
```diff
@@ -663,10 +663,26 @@ def __apply_private_op(
...
+            if self.session.protocol.share_class == ShareTensor:
+
+                result = spdz.mul_master(self, y, op_str, kwargs_)
+                result.shape = MPCTensor._get_shape(op_str, self.shape, y.shape)
+
+            elif self.session.protocol.share_class == ReplicatedSharedTensor:
+
+                result = Falcon.mul(self, y, self.session)
+                result = MPCTensor(
+                    shares=result, shape=self.shape, session=self.session
```
rasswanth-s, line 679: MAYBE : for shape you could use `MPCTensor._get_shape` as it changes during the operation.

src/sympc/tensor/mpc_tensor.py
```diff
@@ -736,6 +753,34 @@ def _get_shape(
         res = op(x, y, **kwargs_)
         return res.shape
 
+    def truncation(
```
rasswanth-s, line 756: Splitting the truncation would be great :+1:

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -306,63 +306,97 @@ def rsub(
         """
         return self.__apply_op(y, "sub")
 
-    def mul(self, y):
+    def mul(self, y: Union[int, float]) -> "ReplicatedSharedTensor":
```
rasswanth-s, line 309:
```suggestion
    def mul(self, y: Union[int, float,torch.Tensor]) -> "ReplicatedSharedTensor":
```

src/sympc/tensor/mpc_tensor.py
kamathhrishi: You mean I should split? or it's good that I split it.

src/sympc/tensor/mpc_tensor.py
kamathhrishi: result._get_shape?

rasswanth-s: Something like this:
https://github.com/OpenMined/SyMPC/blob/2005f3393c03296bbdf18cecafc9c0582783e84f/src/sympc/tensor/mpc_tensor.py#L769

src/sympc/protocol/falcon/falcon.py
kamathhrishi: Okay, I will be implementing it since we don't have to change code when we move P2P someday.

src/sympc/tensor/mpc_tensor.py
kamathhrishi: Alright then. Good catch.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +64,100 @@ def __eq__(self, other: Any) -> bool:
             return False
 
         return True
+
+    @staticmethod
+    def mul_master(x, y, session) -> ReplicatedSharedTensor:
```
rasswanth-s, line 69: Could you add the type annotations in function for `x`,`y`,`session` definition similar to the docstring.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +64,100 @@ def __eq__(self, other: Any) -> bool:
...
+            y (MPCTensor): Another secret
+
+        Returns:
+            shares (ReplicatedSharedTensor): Shares in terms of ReplicatedSharedTensor.
+        """
+        z_shares = []
+
+        for party in range(0, len(x.share_ptrs)):
+
+            z_value = (
```
rasswanth-s, line 155: MAYBE : Can we delegate the multiplication functionality to the respective tensor, similar like what we do in `ShareTensor`, as like the parties could do the functionalities, the orchestrator could act only a control center, this would be helpful in P2P setting also.So that we can realize the functionality from tensor level itself.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +64,100 @@ def __eq__(self, other: Any) -> bool:
...
+
+        # Add random mask to vals
+        shares = [
+            result[0].get() + random0,
+            result[1].get() + random1,
+            result[2].get() + random2,
+        ]
+
+        shares = ReplicatedSharedTensor.distribute_shares(shares, x.session)
+        return shares
```
rasswanth-s, line 136: MAYBE : we could return MPCTensor here itself, similar to what `spdz mul_master` does, just to maintain the convention.

rasswanth-s (approved):

aanurraj (approved):
Looks good left few comments :D

src/sympc/tensor/mpc_tensor.py
```diff
@@ -663,10 +664,32 @@ def __apply_private_op(
...
+                        f"{op_str} has not implemented for ReplicatedSharedTensor"
+                    )
+
+            else:
+
+                raise TypeError("Invalid Share Class")
 
-            result = spdz.mul_master(self, y, op_str, kwargs_)
-            result.shape = MPCTensor._get_shape(op_str, self.shape, y.shape)
         elif op_str in {"sub", "add"}:
```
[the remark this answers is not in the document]
kamathhrishi: Right. This is for a future PR? or if not how do I proceed?

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -26,3 +30,89 @@ def test_exception_malicious_less_parties(get_clients, parties=2) -> None:
...
+    tensor1 = MPCTensor(secret=secret1, session=session)
+    tensor2 = MPCTensor(secret=secret2, session=session)
+
+    result = tensor1 * tensor2
+
+    assert (result.reconstruct() == (secret1 * secret2)).all()
+
+
+@pytest.mark.parametrize("parties", [2, 4])
+def test_mul_private_exception_nothreeparties(get_clients, parties):
```
gmuraru, line 80: MAYBE: Also parametrize this with ```security_type``` as you did in the other tests

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -26,3 +30,89 @@ def test_exception_malicious_less_parties(get_clients, parties=2) -> None:
 def test_invalid_security_type():
     with pytest.raises(ValueError):
         Falcon(security_type="covert")
+
+
+def test_mul_private_integer(get_clients):
```
gmuraru, line 35: MAYBE: Also parametrize this

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -247,3 +252,43 @@ def test_ops_share_public(op_str, precision, base) -> None:
     tensor_decoded = res.fp_encoder.decode(res.shares[0])
 
     assert np.allclose(tensor_decoded, expected_res, rtol=base ** -precision)
+
+
+@pytest.mark.parametrize("parties", [3, 5, 7])
+@pytest.mark.parametrize("security", ["malicious", "semi-honest"])
+def test_ops_publicmul_integer(get_clients, parties, security):
```
gmuraru, line 259: ```public_mul``` not ```publicmul``` to have a consistent name across our tests

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -247,3 +252,43 @@ def test_ops_share_public(op_str, precision, base) -> None:
...
+
+    tensor = MPCTensor(secret=secret, session=session)
+    result = tensor * value
+
+    assert (result.reconstruct() == (secret * value)).all()
+
+
+@pytest.mark.parametrize("parties", [3, 5, 7])
+@pytest.mark.parametrize("security", ["malicious", "semi-honest"])
+def test_ops_publicmul_integer_matrix(get_clients, parties, security):
```
gmuraru, line 279: Also here

gmuraru (changes requested):
Check comments

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +65,87 @@ def __eq__(self, other: Any) -> bool:
...
+            )
+        else:
+            raise NotImplementedError(
+                f"mult operation not implemented for {session.protocol.security_type} setting"
+            )
+
+        przs_masks = [
+            session.session_ptrs[index]
+            .przs_generate_random_share(shape=x.shape)
+            .get_shares()
```
gmuraru, line 110: MAYBE: Can't we do first `[0]` and then `get`? (it might be faster ?)

tests/sympc/protocol/falcon/falcon_test.py
kamathhrishi: Currently, we support only semi-honest protocol for 3 parties so test for the same.

src/sympc/tensor/mpc_tensor.py
```diff
@@ -736,6 +756,34 @@ def _get_shape(
         res = op(x, y, **kwargs_)
         return res.shape
 
+    def truncation(
+        self, result: "MPCTensor", op_str: str, is_private: bool
```
gmuraru, line 760: MAYBE: This should be changed to another name and assign `result` inside this function

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -123,14 +123,12 @@ def get_shares(self) -> List[torch.Tensor]:
     def sanity_checks(
         x: "ReplicatedSharedTensor",
         y: Union[int, float, torch.Tensor, "ReplicatedSharedTensor"],
-        op_str: str,
-    ) -> "ReplicatedSharedTensor":
+    ) -> tuple("ReplicatedSharedTensor",):
```
gmuraru, line 126: MAYBE: `python3.8` does not have this feature to use basic types as type hints.
Use `Tuple` from the `typing` library

gmuraru: + Add all the results

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -174,19 +190,9 @@ def __apply_public_op(
         Raises:
             ValueError: If "op_str" is not supported.
         """
-        y = ReplicatedSharedTensor.sanity_checks(self, y, op_str)
-        rank = 0
-        nr_parties: int = 1  # to handle the case when session_uuid is none.
+        y, other = ReplicatedSharedTensor.sanity_checks(self, y)
```
gmuraru, line 193: MAYBE: ```other``` is not a good name for this

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -218,15 +224,9 @@ def __apply_private_op(
         Raises:
             ValueError: If "op_str" not supported.
         """
-        y = ReplicatedSharedTensor.sanity_checks(self, y, op_str)
+        y, other = ReplicatedSharedTensor.sanity_checks(self, y)
```
gmuraru, line 227: MAYBE: Neither here

tests/sympc/protocol/falcon/falcon_test.py
kamathhrishi: Only support semi-honest so yeah.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -306,63 +306,97 @@ def rsub(
...
         """Apply the "mul" operation between "self" and "y".
 
         Args:
             y: self*y
 
+        Returns:
+            ReplicatedSharedTensor: Result of the operation.
+
         """
+        y, other = self.sanity_checks(self, y)
```
gmuraru, line 319: MAYBE: Neither here

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -306,63 +306,97 @@ def rsub(
...
+
+        shares = []
```
gmuraru, line 321: MAYBE: List comprehension:
```
shares = [share * y.shares[0] for share in shares]
```

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +65,87 @@ def __eq__(self, other: Any) -> bool:
...
+            y (MPCTensor): Another secret
+
+        Returns:
+            shares (ReplicatedSharedTensor): Shares in terms of ReplicatedSharedTensor.
+        """
+        z_shares = []
+
+        for party in range(0, len(x.share_ptrs)):
+
+            z_value = (
```
gmuraru, line 143: Q: can't you use ```parallel_execution``` for this (also for the PRZS) from the caller. In that case you won't have that much communcation

src/sympc/protocol/falcon/falcon.py
kamathhrishi: Yeah, I had tried this first. 
przs_generate_random_share returns Anypointer which isn't subscriptable throwing an error:

`FAILED falcon_test.py::test_mul_private_integer - TypeError: 'AnyPointer' object is not subscriptable`

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +66,84 @@ def __eq__(self, other: Any) -> bool:
...
+        """Implementation of Falcon's multiplication with semi-honest security guarantee.
+
+        Args:
+            x (ReplicatedSharedTensor): Secret
+            y (ReplicatedSharedTensor): Another secret
+
+        Returns:
+            shares (ReplicatedSharedTensor): results in terms of ReplicatedSharedTensor.
+        """
+        z_value = (
```
rasswanth-s, line 144: MAYBE : we could perform the multiplication,in RSTensor itself,since we are calling the falcon protocol again. Something Like 
<pre><code>op = getattr(operator, "mul")
for x_share, y_share in zip(x.share_ptrs, y.share_ptrs):
z_shares.append(op(x_share, y_share))</code></pre>
we could do it locally , it would return a RSTensor pointer,so we could perform przs masking with round of communication(`get`)

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -306,63 +307,104 @@ def rsub(
...
+            result = [Falcon.multiplication_protocol(self, y)]
+
+        else:
+
+            result = [share * y.shares[0] for share in self.shares]
+
+        tensor = ReplicatedSharedTensor(
+            ring_size=self.ring_size, session_uuid=self.session_uuid, config=self.config
+        )
+        tensor.shares = result
```
rasswanth-s, line 337: I think here we could modify , if we incorporate the falcon change as only `return self.__apply_op(y,"mul")`

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -123,14 +124,12 @@ def get_shares(self) -> List[torch.Tensor]:
     def sanity_checks(
         x: "ReplicatedSharedTensor",
         y: Union[int, float, torch.Tensor, "ReplicatedSharedTensor"],
-        op_str: str,
-    ) -> "ReplicatedSharedTensor":
+    ) -> Tuple["ReplicatedSharedTensor", Tuple[Session, int, Config, int, int]]:
```
rasswanth-s, line 127: MAYBE: we could have another argument , something like boolean value named `parties_check` or `falcon_check` for ops like `mul`, `matmul` , if set to true, it will check ,if `number of parties !=3` , if session_uuid id not set, we could have `len(shares)!=2`, for sanity checks. The boolean value will default value `False`.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -218,15 +225,9 @@ def __apply_private_op(
         Raises:
             ValueError: If "op_str" not supported.
         """
-        y = ReplicatedSharedTensor.sanity_checks(self, y, op_str)
+        y, session_vals = ReplicatedSharedTensor.sanity_checks(self, y)
```
rasswanth-s, line 228: MAYBE : Here , we could have , if we have  op_str in `{mul,matmul` we could pass `sanity_checks(...,falcon_check=True)`

rasswanth-s: `falcon_check` or `parties_check` or some name which suits best for the functionality.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Apply_op becomes crowded with lots of if-else statements. sharetensor it makes sense because they outsource it to pytorch.

src/sympc/protocol/falcon/falcon.py
kamathhrishi: Yeah, I wll incorporate this.

kamathhrishi: I tried this, but then RSTensor is not able to access the actual session. The session it accessed has 0 parties so the shares are not being generated. I will try it for a while if I don't find a neat way out. I will create an issue for optimizing this which we can solve next.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Sanity check is native to RSTensor so would be good to keep protocol related information separate or wherever it comes into play (mul for 3 parties)

src/sympc/protocol/falcon/falcon.py
kamathhrishi: This was also the reason I had to fall back to checking number of shares rather than number of parties in session.

rasswanth-s:
Looks good to me. :+1:

src/sympc/tensor/mpc_tensor.py
```diff
@@ -736,6 +758,37 @@ def _get_shape(
...
+            input_tensor (MPCTensor): Result of operation
+            op_str (str): Operation name
+            is_private (bool): If operation is private
+
+        Returns:
+            result (MPCTensor): Truncated result
+        """
+        # Truncation should be added for RSTensor
+        if (
+            op_str in {"mul", "matmul", "conv2d", "conv_transpose2d"}
```
Syzygianinfern0, line 776: Could we keep this set in a more accessible place as we might be adding more to it?

Something like these: 

https://github.com/OpenMined/SyMPC/blob/94fd2f3a440d0bad4a289a1fdece9b436fa34d5a/src/sympc/tensor/mpc_tensor.py#L29

Syzygianinfern0 (approved):
LGTM! :100: 
Great addition to our codebase! [or should I say multiplication? :stuck_out_tongue: (excuse the cringe)]

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +66,84 @@ def __eq__(self, other: Any) -> bool:
...
+        if len(session.parties) != 3:
+            raise ValueError("Falcon requires 3 parties")
+
+        result = None
+
+        if session.protocol.security_type == "semi-honest":
+
+            z_shares = []
+            for party in range(0, len(x.share_ptrs)):
+                z_shares.append(x.share_ptrs[party] * y.share_ptrs[party])
```
gmuraru, line 100: MAYBE: Also this should be in the parallel execution.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +66,84 @@ def __eq__(self, other: Any) -> bool:
...
+                share = z_shares[party_rank].get_shares().get()[0] + przs_mask
+                return share
+
+            shares = parallel_execution(get_shares_and_add_mask)([[0], [1], [2]])
+            reshared_shares = ReplicatedSharedTensor.distribute_shares(
+                shares, x.session
+            )
+            result = MPCTensor(shares=reshared_shares, session=x.session)
+
+        elif session.protocol.security_type == "malicious":
```
gmuraru, line 120: MAYBE: Remove this since it would go on the else case

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -306,63 +307,111 @@ def rsub(
...
+        Raises:
+            ValueError: Raised when private mul is performed parties!=3.
+
         """
+        y_tensor, session_vals = self.sanity_checks(self, y)
+        session, ring_size, config, rank, nr_parties = session_vals
+        is_private = isinstance(y, ReplicatedSharedTensor)
+
+        if is_private:
+            # currently there is no way of getting session of RStensor since session is optional.
```
gmuraru, line 328: TYPO: ```RSTTEnsor```

gmuraru: There should be a way to get it, right? (because you serialize the ```session_uuid``` and you get that?)

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -157,7 +156,25 @@ def sanity_checks(
...
+            session = sympc.session.get_session(str(x.session_uuid))
+            ring_size = session.ring_size
+            config = session.config
+            rank = session.rank
+            nr_parties = session.nr_parties
+        else:
+            ring_size = x.ring_size
+            config = x.config
+
+        return y, (session, ring_size, config, rank, nr_parties)
```
gmuraru, line 177: MAYBE: Why don't you simply return the session (in case there is no session you create one with the specific values)

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Oh, I have to remove the comment. I did get the session and check for a number of parties.

src/sympc/tensor/mpc_tensor.py
kamathhrishi: Hmmm,  I am not sure if adding it to an accessible place would help since this is the only function we will be using it in, and can't think of a common name for grouping them. But, in general, I think this whole MPCclass needs refactoring in a future PR.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Hmmm, but how do we know the number of parties?

src/sympc/protocol/falcon/falcon.py
kamathhrishi: Done

src/sympc/protocol/falcon/falcon.py
kamathhrishi: Done

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +66,81 @@ def __eq__(self, other: Any) -> bool:
...
+            raise ValueError("Falcon requires 3 parties")
+
+        result = None
+
+        if session.protocol.security_type == "semi-honest":
+
+            def multiply_shares(party_rank):
+                return x.share_ptrs[party_rank] * y.share_ptrs[party_rank]
+
+            z_shares = parallel_execution(multiply_shares)([[0], [1], [2]])
```
gmuraru, line 101: OUGHTTO: my question is more -- why don't you have only one `parallel_execution` with both the functionaltities?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +66,81 @@ def __eq__(self, other: Any) -> bool:
...
+                przs_mask = (
+                    session.session_ptrs[party_rank]
+                    .przs_generate_random_share(shape=x.shape)
+                    .get_shares()
+                    .get()[0]
+                )
+                share = z_shares[party_rank].get_shares().get()[0] + przs_mask
+                return share
+
+            shares = parallel_execution(get_shares_and_add_mask)([[0], [1], [2]])
```
gmuraru, line 115: MAYBE: The ranks can be obtained from the session using ```get_session```

src/sympc/tensor/replicatedshare_tensor.py
gmuraru: So this function it is used in the case we are in the MPC setup - we have the `session_uuid` and not in the MPC Setup - the `session_uuid` is `None` and in that scenario, we do not have `session`, right?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -522,18 +574,33 @@ def distribute_shares(shares: List[ShareTensor], session: Session) -> List:
...
 
+        Raises:
+            TypeError: when Datatype of shares is invalid.
+
         """
-        parties = session.parties
+        if not isinstance(shares, list):
+            raise TypeError("Shares to be distributed should be a list of shares")
 
+        if len(shares) == 0:
```
gmuraru, line 584: MAYBE: Add an exception in the case where `len(shares)` is different than `nr_parties`

gmuraru (changes requested):
See comments

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +69,97 @@ def __eq__(self, other: Any) -> bool:
...
+
+        else:
+            raise NotImplementedError(
+                f"mult operation not implemented for {session.protocol.security_type} setting"
+            )
+
+        return result
+
+    @staticmethod
+    def compute_zvalue_and_add_mask(x, y, przs_mask) -> torch.Tensor:
```
rasswanth-s, line 130: MAYBE: we could add type annotations for the arguments.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +69,103 @@ def __eq__(self, other: Any) -> bool:
...
+        result = None
+
+        if session.protocol.security_type == "semi-honest":
+            args = []
+            for index in range(0, 3):
+                args.append(
+                    [
+                        x.share_ptrs[index],
+                        y.share_ptrs[index],
+                        session.session_ptrs[index],
```
gmuraru, line 106: MAYBE: You don't need to send the session - you can send only ```x and y``` (as you already do) - and in  ```Falcon.compute_zvalue_and_add_mask``` do:
```
session_uuid = x.session_uuid
session = get_session(str(session_uuid)) # get_session is in session_utils.py
```

src/sympc/protocol/falcon/falcon.py
```diff
@@ -63,3 +69,103 @@ def __eq__(self, other: Any) -> bool:
...
+                    ]
+                )
+            z_shares_ptrs = parallel_execution(
+                Falcon.compute_zvalue_and_add_mask, session.parties
+            )(args)
+
+            z_shares = []
```
gmuraru, line 113: MAYBE:
```
z_shares = [share.get() for share in z_shares_ptrs]
```

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -306,63 +301,108 @@ def rsub(
...
+
         """
+        y_tensor, session = self.sanity_checks(self, y)
+        is_private = isinstance(y, ReplicatedSharedTensor)
+
+        if is_private:
+            if session.nr_parties == 3:
+                from sympc.protocol import Falcon
+
+                result = [Falcon.multiplication_protocol(self, y_tensor)]
```
gmuraru, line 324: Q: Why not?
```
self.session.protocol.multiplication_protocol(self, y_tensor)
```

gmuraru: In that case you won't need to import anything

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -157,7 +158,16 @@ def sanity_checks(
...
 
-        return y
+        session_uuid = x.session_uuid
+
+        if session_uuid is not None:
+            session = sympc.session.get_session(str(x.session_uuid))
+        else:
+            ring_size = x.ring_size
+            config = x.config
+            session = Session(config=config, ring_size=ring_size, parties=["alice"])
```
gmuraru, line 168: BUG: parties["alice"] should be removed, right?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -522,18 +567,37 @@ def distribute_shares(shares: List[ShareTensor], session: Session) -> List:
...
 
+        Raises:
+            TypeError: when Datatype of shares is invalid.
+
         """
-        parties = session.parties
+        if not isinstance(shares, list):
+            raise TypeError("Shares to be distributed should be a list of shares")
+
+        if len(shares) == 0:
```
gmuraru, line 577: MAYBE: Remove this, since this case would be already covered by the bellow code, right?

src/sympc/api.py
```diff
@@ -175,6 +176,7 @@
         "sympc.tensor.ReplicatedSharedTensor.clone",
         "sympc.tensor.ReplicatedSharedTensor",
     ),
+    ("sympc.tensor.ReplicatedSharedTensor.shape", "syft.lib.python.Tuple"),
```
gmuraru, line 179: Q: I think you can remove this, right? (since now you call shape from a "local" context)

src/sympc/tensor/mpc_tensor.py
```diff
@@ -768,15 +817,8 @@ def __apply_op(
...
-            not is_private or self.session.nr_parties > 2
-        ):
-            # For private op we do the division in the mul_parties function from spdz
-            scale = (
-                self.session.config.encoder_base
-                ** self.session.config.encoder_precision
-            )
-            result = result.truediv(scale)
+        # Check operation and apply truncation if required.
+        result = self.truncation(result, op_str, is_private)
```
gmuraru, line 821: NITPICK: The method name should be a verb, not a noun. Could we rename this to ```self.truncate```?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -123,14 +126,12 @@ def get_shares(self) -> List[torch.Tensor]:
     def sanity_checks(
         x: "ReplicatedSharedTensor",
         y: Union[int, float, torch.Tensor, "ReplicatedSharedTensor"],
-        op_str: str,
-    ) -> "ReplicatedSharedTensor":
+    ) -> Tuple["ReplicatedSharedTensor", Tuple[Session, int, Config, int, int]]:
```
gmuraru, line 129: MAYBE: This is not up-to-date

src/sympc/api.py
kamathhrishi: Nope, this was actually the fix for the error where ReplicatedShareTensor was being passed as a shape to PRZS. Also in the future, we might need it anyway?

gmuraru: Yep, but we should add it when we need it.
I am not sure if now we are using it. If you remove it, it should still work (this allowlist is for being able to call shape on a remote RST)

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -43,7 +43,7 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
 
     # Used by the SyMPCTensor metaclass
     METHODS_FORWARD = {"numel", "t", "unsqueeze", "view", "sum", "clone"}
-    PROPERTIES_FORWARD = {"T"}
+    PROPERTIES_FORWARD = {"T", "shape"}
```
gmuraru, line 46: We might need to also remove it from here. We could add it when we would need it.

gmuraru (approved):
Only some comments. In rest, it looks good to me

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: I need this because when generating a PRZS share, it should be the shape of RSTensor.

src/sympc/api.py
kamathhrishi: Yeah, dint need this.

PR #253, opened 2021-06-24
Feat/ci scheduled tests

### LINKED ISSUE #244: Modification of CI to run scheduled tests.
## Description
Our current CI  runs all the test suite, which makes it slow.We would ideally want to run limited number of github runners,similar like PySyft which runs only latest versions of python for its tests,and to also integrate scheduled CI tests everyday which runs the whole test suite.This would greatly improve our CI.

## Additional Context
Syft PR Tests : [PR_Tests](https://github.com/OpenMined/PySyft/blob/dev/.github/workflows/syft-pr_tests.yml)
Schduled Tests : [Scheduled_Tests](https://github.com/OpenMined/PySyft/blob/dev/.github/workflows/syft-version_tests.yml)

danielorihuela:
## Description
Closes #244 

## How has this been tested?
Pushing experiments in my fork.

## Checklist

### DIFF
--- a/.github/workflows/scheduled_tests.yml
+++ b/.github/workflows/scheduled_tests.yml
@@ -1,0 +1,51 @@
+name: Scheduled Tests
+
+on:
+  schedule:
+    - cron: "0 22 * * *" # 7pm Brazil, 10pm UTC, 8am AEST
+
+jobs:
+  python-tests-all-versions:
+    strategy:
+      max-parallel: 9
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: [3.7, 3.8, 3.9]
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Upgrade pip
+        run: |
+          pip install --upgrade --user pip
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+      - name: pip cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-py${{ matrix.python-version }}-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-py${{ matrix.python-version }}-
+      - name: Install CPU Torch
+        if: runner.os != 'macos'
+        run: |
+          pip install torch==1.8.1+cpu torchvision==0.9.1+cpu torchcsprng==0.2.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
+        shell: bash
+      - name: Install dependencies
+        run: |
+          hash -r
+          pip install -r requirements.dev.txt -f https://download.pytorch.org/whl/torch_stable.html
+          pip install -e .
+        shell: bash
+
+      - name: Run tests
+        run: |
+          pytest tests/ --deselect=tests/notebooks -n 8 -x
--- a/.github/workflows/scheduled_tests.yml
+++ b/.github/workflows/scheduled_tests.yml
@@ -2,7 +2,7 @@ name: Scheduled Tests
 
 on:
   schedule:
-    - cron: "0 22 * * *" # 7pm Brazil, 10pm UTC, 8am AEST
+    - cron: "0 2 * * *" # 2am UTC
 
 jobs:
   python-tests-all-versions:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,10 +24,10 @@ on:
 jobs:
   python-tests:
     strategy:
-      max-parallel: 9
+      max-parallel: 3
       matrix:
-        os: [ubuntu-latest, macos-latest]
-        python-version: [3.7, 3.8, 3.9]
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -24,10 +24,10 @@ on:
 jobs:
   python-tests:
     strategy:
-      max-parallel: 9
+      max-parallel: 3
       matrix:
-        os: [ubuntu-latest, macos-latest]
-        python-version: [3.7, 3.8, 3.9]
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
rasswanth-s, line 29:
```suggestion
        os: [ubuntu-latest, macos-latest]
```
I think we could remove windows from our CI tests.

.github/workflows/tests.yml
```diff
@@ -24,10 +24,10 @@ on:
...
+        python-version: [3.9]
```
rasswanth-s, line 30: Maybe we could also have a linting check run first like PySyft and then run other runners if it passes , so we could use our runners optimally.

rasswanth-s: https://github.com/OpenMined/PySyft/blob/563d00f22c7577fcd616b43397737eac09ed6204/.github/workflows/syft-pr_tests.yml#L25

.github/workflows/scheduled_tests.yml
```diff
@@ -0,0 +1,51 @@
...
+on:
+  schedule:
+    - cron: "0 22 * * *" # 7pm Brazil, 10pm UTC, 8am AEST
+
+jobs:
+  python-tests-all-versions:
+    strategy:
+      max-parallel: 9
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
rasswanth-s, line 12:
```suggestion
        os: [ubuntu-latest, macos-latest]
```
I think we could also remove windows runner here.

rasswanth-s:
Currently when our CI tests run , if one of the runners fails, other passed, we would have to run `re-run all jobs` which re runs all the runners , including the one which passed before in that particular `workflow` file.This is a long standing issue in github runners,where we can't restart a single runner in a workflow if it fails.
One solution we split PR workflow as `ubuntu-tests.yml` `mac-os-test.yml`  , so we can run the specific runner if it fails,instead running as a whole. @gmuraru Shall we do it like this?

.github/workflows/scheduled_tests.yml
danielorihuela: I am not sure about that. If we remove it from here, SyMPC will never get tested on windows.

gmuraru: I think we should also test the windows node here - since those jobs should not "bother" the devs. Could we put those at an hour like 2:00 AM UTC?

rasswanth-s: The reason it was removed was that ,it was very slow on the CI tests and failed frequently,so it was removed.
Maybe we could test it,when we are releasing a package or stable release @gmuraru

rasswanth-s: Yeah at schduled_tests we could have it

gmuraru:
> Currently when our CI tests run , if one of the runners fails, others passed, we would have to run `re-run all jobs` which re runs all the runners , including the one which passed before in that particular `workflow` file.This is a long standing issue in github runners,where we can't restart a single runner in a workflow if it fails.
> 
> One solution we split PR workflow as `ubuntu-tests.yml` `mac-os-test.yml` , so we can run the specific runner if it fails,instead running as a whole. @gmuraru Shall we do it like this?

Hmmm...I do not think if we want this for the night tests (to rerun them if they fail). If they fail we should investigate what failed and create tickets to consolidate the codebase such that they won't fail anymore.

LaRiffle (approved):

gmuraru (approved):
LGTM!

rasswanth-s:
> > Currently when our CI tests run , if one of the runners fails, others passed, we would have to run `re-run all jobs` which re runs all the runners , including the one which passed before in that particular `workflow` file.This is a long standing issue in github runners,where we can't restart a single runner in a workflow if it fails.
> > One solution we split PR workflow as `ubuntu-tests.yml` `mac-os-test.yml` , so we can run the specific runner if it fails,instead running as a whole. @gmuraru Shall we do it like this?
> 
> Hmmm...I do not think if we want this for the night tests (to rerun them if they fail). If they fail we should investigate what failed and create tickets to consolidate the codebase such that they won't fail anymore.

Yeah george for the nightly tests,we could keep  as it, so we can rectify the error, I meant for the PR CI tests, where we rerun the whole workflow many times if one fails.

gmuraru:
> > > Currently when our CI tests run , if one of the runners fails, others passed, we would have to run `re-run all jobs` which re runs all the runners , including the one which passed before in that particular `workflow` file.This is a long standing issue in github runners,where we can't restart a single runner in a workflow if it fails.
> > > One solution we split PR workflow as `ubuntu-tests.yml` `mac-os-test.yml` , so we can run the specific runner if it fails,instead running as a whole. @gmuraru Shall we do it like this?
> > 
> > 
> > Hmmm...I do not think if we want this for the night tests (to rerun them if they fail). If they fail we should investigate what failed and create tickets to consolidate the codebase such that they won't fail anymore.
> 
> Yeah george for the nightly tests,we could keep as it, so we can rectify the error, I meant for the PR CI tests, where we rerun the whole workflow many times if one fails.

Hmm...even for the other I would tend to keep them together only that we don't diverge at one point - add some options on Linux on forgetting to add them on OSX. I think that it might be better to start fixing the flaky tests than split those up at the moment - also the tests should fail pretty fast and we can rerun them if there is any problem.

rasswanth-s (approved):
LGTM

PR #254, opened 2021-06-25
Fix gradient not being recorded

gmuraru:
## Description
Track also the operations from static using the `forward` method call.
We tracked only the operations we did on ```MPCTensor```, but we need to be able to also track the operations from ```functional```.

## Affected Dependencies
The `relu` and `maxpool_2d` grad.

## How has this been tested?
- The tests should pass as they are

## Checklist

### DIFF
--- a/src/sympc/grads/grad_functions.py
+++ b/src/sympc/grads/grad_functions.py
@@ -636,6 +636,7 @@ class GradMaxPool2D(GradFunc):
         stride: Optional[Union[int, Tuple[int, int]]] = None,
         padding: Union[int, Tuple[int, int]] = 0,
         dilation: Union[int, Tuple[int, int]] = 1,
+        return_indices: bool = False,
     ) -> MPCTensor:
         """Perform the feedforward and compute the result for the MaxPool2D operation.
 
@@ -650,6 +651,7 @@ class GradMaxPool2D(GradFunc):
                 in case it is passed as an integer then that value is used for height and width
             dilation (Union[int, Tuple[int, int]]): the dilation size
                 in case it is passed as an integer then that value is used for height and width
+            return_indices (bool): to return the indices of the max values
 
         Returns:
             res (MPCTensor): The result of the reshape operation
--- a/src/sympc/grads/grad_functions.py
+++ b/src/sympc/grads/grad_functions.py
@@ -562,19 +562,21 @@ class GradReshape(GradFunc):
     """The Reshape gradient function."""
 
     @staticmethod
-    def forward(ctx: Dict[str, Any], x: MPCTensor, shape: Tuple[int]) -> MPCTensor:
+    def forward(
+        ctx: Dict[str, Any], x: MPCTensor, *shape: Union[List[int], Tuple[int]]
+    ) -> MPCTensor:
         """Perform the feedforward and compute the result for the reshape operation.
 
         Args:
             ctx (Dict[str, Any]): Context used to save information needed in the backward pass
             x (MPCTensor): the MPCTensor to be reshaped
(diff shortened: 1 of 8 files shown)


src/sympc/tensor/mpc_tensor.py
```diff
@@ -865,21 +865,21 @@ def __getattribute__(self, attr_name: str) -> Any:
...
         session = object.__getattribute__(self, "session")
         if grad_fn and session.autograd_active:
-            from sympc.tensor.grads import forward
+            from sympc.grads import forward
 
             return functools.partial(forward, self, grad_fn)
 
+        if attr_name in STATIC_FUNCS.keys():
+            return functools.partial(STATIC_FUNCS[attr_name], self)
+
```
Syzygianinfern0, line 882: Why this move?

src/sympc/grads/grad_functions.py
```diff
@@ -636,6 +636,7 @@ def forward(
         stride: Optional[Union[int, Tuple[int, int]]] = None,
         padding: Union[int, Tuple[int, int]] = 0,
         dilation: Union[int, Tuple[int, int]] = 1,
+        return_indices: bool = False,
```
Syzygianinfern0, line 639: Missed out forwarding this arg to the call below (L679)

https://github.com/OpenMined/SyMPC/blob/969718db81690b674a642b29984074fd45286034/src/sympc/grads/grad_functions.py#L673-L680

Syzygianinfern0 (changes requested):

src/sympc/tensor/mpc_tensor.py
gmuraru: I was thinking that the Grad should be checked first and then the statics --- only like an "importance hierarchy" -- for the moment I do not think we have any grad recording in the static file.

src/sympc/grads/grad_functions.py
gmuraru: I should have had an if...else statement - the return indices would be needed if we record the gradient.

tests/sympc/grad/grad_functions_test.py
```diff
@@ -7,23 +7,23 @@
 import torch
 
 import sympc
+from sympc.grads.grad_functions import GradAdd
```
gmuraru, line 10: This was done by ```isort```

src/sympc/module/nn/functional.py
```diff
@@ -24,6 +26,9 @@ def relu(x: MPCTensor) -> MPCTensor:
     Returns:
         An MPCTensor which represents the ReLu applied on the input tensor
     """
+    relu_forward = GRAD_FUNCS.get("relu", None)
+    if relu_forward and x.session.autograd_active:
+        return forward(x, relu_forward)
```
aanurraj, line 31: this will the issue for not recording the grad right? but I wonder where is it being stored as node ?

aanurraj (approved):
Thank you, just one more question have you ran the notebook which I gave? it is working on that architecture?

src/sympc/module/nn/functional.py
gmuraru: It is recorded - the forward function deals with this.

gmuraru:
> Thank you, just one more question have you ran the notebook which I gave? it is working on that architecture?

I tested it on the code you gave me on slack - but thank you for remembering me that I did not uncomment some of the lines.

gmuraru:
> Thank you, just one more question have you ran the notebook which I gave? it is working on that architecture?

Good call @aanurraj. I have found a problem with the `reshape` and I fixed it ^_^

PR #255, opened 2021-06-25
Refactoring RSTensor Share Distribution and Optimised unit tests 

kamathhrishi:
## Description
This PR parallelizes RSTensor share distribution and optimizes unit tests by running for just enough parties to test functionality. 

## Checklist

### DIFF
--- a/src/sympc/protocol/falcon/falcon.py
+++ b/src/sympc/protocol/falcon/falcon.py
@@ -98,7 +98,7 @@ class Falcon(metaclass=Protocol):
         result = None
 
         if session.protocol.security_type == "semi-honest":
-            args = [[x.share_ptrs[index], y.share_ptrs[index]] for index in range(3)]
+            args = list(zip(x.share_ptrs, y.share_ptrs))
 
             z_shares_ptrs = parallel_execution(
                 Falcon.compute_zvalue_and_add_mask, session.parties
--- a/src/sympc/protocol/falcon/falcon.py
+++ b/src/sympc/protocol/falcon/falcon.py
@@ -98,18 +98,11 @@ class Falcon(metaclass=Protocol):
         result = None
 
         if session.protocol.security_type == "semi-honest":
-            args = []
-            for index in range(0, 3):
-                args.append(
-                    [
-                        x.share_ptrs[index],
-                        y.share_ptrs[index],
-                    ]
-                )
+            args = [[x.share_ptrs[index], y.share_ptrs[index]] for index in range(3)]
+
             z_shares_ptrs = parallel_execution(
                 Falcon.compute_zvalue_and_add_mask, session.parties
             )(args)
-
             z_shares = [share.get() for share in z_shares_ptrs]
 
             # Convert 3-3 shares to 2-3 shares by resharing
--- a/src/sympc/tensor/replicatedshare_tensor.py
+++ b/src/sympc/tensor/replicatedshare_tensor.py
@@ -78,9 +78,7 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
         self.shares = []
 
         if shares is not None:
-            self.shares = [
-                self._encode(shares[i]).to(tensor_type) for i in range(len(shares))
-            ]
+            self.shares = [self._encode(share).to(tensor_type) for share in shares]
 
     def _encode(self, data: torch.Tensor) -> torch.Tensor:
         """Encode via FixedPointEncoder.
@@ -110,9 +108,10 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
         """
         shares = []
 
-        for share in self.shares:
-            shares.append(self.fp_encoder.decode(share.type(torch.LongTensor)))
-
+        shares = [
+            self.fp_encoder.decode(share.type(torch.LongTensor))
+            for share in self.shares
+        ]
         return shares
 
     def get_shares(self) -> List[torch.Tensor]:
@@ -675,7 +674,8 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
             shares = []
 
             for share in _self.shares:
-                shares.append(getattr(share, property_name))
+                tensor = getattr(share, property_name)
+                shares.append(tensor)
 
             res = ReplicatedSharedTensor(
                 session_uuid=_self.session_uuid, config=_self.config
@@ -717,7 +717,8 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
         ) -> Any:
             shares = []
             for share in _self.shares:
-                shares.append(getattr(share, method_name)(*args, **kwargs))
+                tensor = getattr(share, method_name)(*args, **kwargs)
+                shares.append(tensor)
 
             res = ReplicatedSharedTensor(
                 session_uuid=_self.session_uuid, config=_self.config
--- a/src/sympc/tensor/replicatedshare_tensor.py
+++ b/src/sympc/tensor/replicatedshare_tensor.py
@@ -582,6 +582,44 @@ class ReplicatedSharedTensor(metaclass=SyMPCTensor):
         raise ValueError("Invalid security Type")
 
     @staticmethod
+    def distribute_shares_to_party(
+        shares: List[Union[ShareTensor, torch.Tensor]],
+        party_rank: int,
+        session: Session,
+    ):
+        """Distributes shares to party.
+
+        Args:
+            shares (List[Union[ShareTensor,torch.Tensor]]): Shares to be distributed.
+            party_rank (int): Rank of party.
+            session (Session): Current session
+
+        Returns:
+            tensor (ReplicatedSharedTensor): Tensor with shares
+        """
+        party = session.parties[party_rank]
+        nshares = session.nr_parties - 1
+        party_shares = []
+
+        for i in range(party_rank, party_rank + nshares):
+            share = shares[i % (nshares + 1)]
+
+            if isinstance(share, torch.Tensor):
+                party_shares.append(share)
+
+            elif isinstance(share, ShareTensor):
+                tensor = share.tensor
+                party_shares.append(tensor)
+
+        tensor = ReplicatedSharedTensor(
+            party_shares,
+            config=Config(encoder_base=1, encoder_precision=0),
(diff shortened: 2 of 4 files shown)


tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -306,7 +306,7 @@ def test_ops_public_mul_integer(get_clients, parties, security):
     assert (result.reconstruct() == (secret * value)).all()
 
 
-@pytest.mark.parametrize("parties", [3, 5, 7])
```
gmuraru, line 309: Q: Those changes are for speed-up?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -581,6 +581,43 @@ def reconstruct(
 
         raise ValueError("Invalid security Type")
 
+    @staticmethod
+    def distribute_shares_to_party(
+        shares: List[Union[ShareTensor, torch.Tensor]],
+        party_rank: int,
+        session: Session,
+    ):
```
LaRiffle, line 589: Add the return annotation type

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -581,6 +581,43 @@ def reconstruct(
...
+        Args:
+            shares (List[Union[ShareTensor,torch.Tensor]]): Shares to be distributed.
+            party_rank (int): Rank of party.
+            session (Session): Current session
+
+        Returns:
+            tensor (ReplicatedSharedTensor): Tensor with shares
+        """
+        party = session.parties[party_rank]
+        nshares = session.nr_parties - 1
```
LaRiffle, line 601: I'd rather not introduce this variable which doesn't seem very useful

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -581,6 +581,43 @@ def reconstruct(
...
+        party_shares = []
+
+        for share_index in range(party_rank, party_rank + nshares):
+            share = shares[share_index % (nshares + 1)]
+
+            if isinstance(share, torch.Tensor):
+                party_shares.append(share)
+
+            elif isinstance(share, ShareTensor):
+                party_shares.append(share.tensor)
```
LaRiffle, line 611: Add a final `else` statement to raise an Typeerror

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -605,32 +642,13 @@ def distribute_shares(
...
-
-            tensor = ReplicatedSharedTensor(
-                party_shares,
-                config=Config(encoder_base=1, encoder_precision=0),
-                session_uuid=session.rank_to_uuid[i],
-            ).send(parties[i])
-            ptr_list.append(tensor)
-
-        return ptr_list
+        return parallel_execution(ReplicatedSharedTensor.distribute_shares_to_party)(
```
LaRiffle, line 649: have you checked that parallel execution is useful here? It creates sub processes that can be slow to launch, so sometimes non parellel (ie classic) execution is faster. I would suggest running X experiments and comparing before implementing this feature

tests/sympc/tensor/mpc_tensor_test.py
```diff
@@ -125,7 +125,7 @@ def test_local_secret_not_tensor(get_clients) -> None:
     assert np.allclose(torch.tensor(x_float), result)
 
 
-@pytest.mark.parametrize("nr_clients", [2, 3, 4, 5])
+@pytest.mark.parametrize("nr_clients", [2, 5])
```
LaRiffle, line 128: Agreed with @gmuraru, why do we change this?

LaRiffle (changes requested):
Some great syntax optimisations! I've left some comments

kamathhrishi:
@gmuraru @LaRiffle Those are changes for speedup. 
I think it wasn't right for me to assume what works for 2 or 5 parties will work the same for 3,4 or 7 parties for FSS and SPDZ.

I noticed that was the case for ReplicatedSharedTensor (RSTensor) operations implemented and removed tests I had written for 7&11 parties. In the case of RSTensor what works for 2 or 5 parties works well for say 3,4,...,n parties so having tests for 7 & 11 parties were piling up CI time and compute. Unit tests should test for correct functionality. 

After removing just tests for 7&11 parties I noticed time taken to run RSTensor tests reduced from ~800s to ~230s. 
MPCTensor from ~5500s to ~2300. In MPCTensors tests, I removed some tests that I thought were redundant like instead of having tests for 2,3,4,5, and just 2 & 5. 

For RStensor having just 2 & 5 parties (some case 3&5 or just 3) tests would work. I am not sure if functionality varies for different parties for SPDZ and FSS.

What do you think?

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Oh alright. Will have a look.

tests/sympc/tensor/mpc_tensor_test.py
kamathhrishi: I have addressed this as a general comment. Let me know what you think.

tests/sympc/tensor/replicatedshare_tensor_test.py
kamathhrishi: Yeah for speed up. I have addressed this as a general comment. Let me know what you think.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Yeah, classic was slightly better, I will switch back to classic.

gmuraru: Yeah we can switch to the older one - I do consider that we might revisit this and do it like you have here.
For the moment we always create a Thread/Process pool for parallelization - I think that if we use 3-4 parties (with VirtualMachines -- the testing setup) it would not give better results because we spend more time creating the thread pool than doing the effective work.

What might change, if we switch to real multi-party over the network, it might be that this approach might give better results depending on the number of parties.

kamathhrishi: I will be keeping the refactoring part. That is having a separate function to distribute shares to a party and an overall function to distribute it to all parties. In the future would just add a parallel_execution statement around it.

tests/sympc/tensor/mpc_tensor_test.py
gmuraru: Let's also add the "3" value for those -- I think that 2, 3, 5 should suffice - 2 for FSS, 3 for Falcon, and 5 for a more general case.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: I feel it makes the code more easily read rather than having just session.nr_parties-1 somewhere, someone reading the code would easily understand its the number of sharea.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -98,18 +98,11 @@ def mul_master(x: MPCTensor, y: MPCTensor, session: Session) -> MPCTensor:
...
         if session.protocol.security_type == "semi-honest":
-            args = []
-            for index in range(0, 3):
-                args.append(
-                    [
-                        x.share_ptrs[index],
-                        y.share_ptrs[index],
-                    ]
-                )
+            args = [[x.share_ptrs[index], y.share_ptrs[index]] for index in range(3)]
```
gmuraru, line 101: MAYBE: `list(zip(x.share_ptrs, y.share_ptrs)`

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -78,8 +78,9 @@ def __init__(
         self.shares = []
 
         if shares is not None:
-            for i in range(len(shares)):
-                self.shares.append(self._encode(shares[i]).to(tensor_type))
+            self.shares = [
+                self._encode(shares[i]).to(tensor_type) for i in range(len(shares))
```
gmuraru, line 82: MAYBE: `self._encode(share).to(tensor_type) for share in shares`

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -109,9 +110,8 @@ def _decode(self) -> List[torch.Tensor]:
         """
         shares = []
 
-        for i in range(len(self.shares)):
-            tensor = self.fp_encoder.decode(self.shares[i].type(torch.LongTensor))
-            shares.append(tensor)
+        for share in self.shares:
```
gmuraru, line 113: Q: Maybe list comprehension also for this?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -695,9 +716,8 @@ def method_new_rs_tensor(
             _self: "ReplicatedSharedTensor", *args: List[Any], **kwargs: Dict[Any, Any]
         ) -> Any:
             shares = []
-            for i in range(len(_self.shares)):
-                tensor = getattr(_self.shares[i], method_name)(*args, **kwargs)
```
gmuraru, line 699: Q: Could we leave them as they were before since it might help  when we try to debug

gmuraru (approved):
Only some comments but in rest, it looks good!

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: Makes sense.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -581,18 +580,61 @@ def reconstruct(
...
+
     @staticmethod
     def distribute_shares(
         shares: List[Union[ShareTensor, torch.Tensor]], session: Session
-    ) -> List:
+    ) -> List["ReplicatedSharedTensor"]:
         """Distribute a list of shares.
 
         Args:
             shares (List[ShareTensor): list of shares to distribute.
```
rasswanth-s, line 633: I think the it should also include torch.Tensor in args.

rasswanth-s (approved):
LGTM!

Issue #172, opened 2021-05-14
FALCON : ABY3 Truncation Protocol


rasswanth-s:
## Feature Description
Implement truncation protocol Π<sub>trunc2</sub> in ABY3.

The FALCON protocol implements it to ensure correct fixed-point precision on multiplication.
Protocol with security guarantees. 
- [x] Malicious 
- [x] Semi honest

## Additional Context
Refer [ABY3](https://eprint.iacr.org/2018/403.pdf)

PR #257, opened 2021-06-26
ABY3 Truncation Protocol -Semihonest

rasswanth-s:
## Description
Implements the ABY3 truncation algorithm in Falcon.
Addresses part of issue #172 

## Affected Dependencies

## Checklist

### DIFF
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -60,10 +60,10 @@ class ABY3(metaclass=Protocol):
         return True
 
     @staticmethod
-    def trunc1(
+    def truncation_algorithm1(
         ptr_list: List[torch.Tensor], shape: torch.Size, session: Session
     ) -> List[ReplicatedSharedTensor]:
-        """Performs the ABY3 trunc1 algorithm.
+        """Performs the ABY3 truncation algorithm1.
 
         Args:
             ptr_list (List[torch.Tensor]): Tensors to truncate
@@ -104,7 +104,7 @@ class ABY3(metaclass=Protocol):
         TODO :Switch to trunc2 algorithm  as it is communication efficient.
         """
         if session.nr_parties != 3:
-            raise ValueError("Share trunc1 algorithm works only for 3 parites.")
+            raise ValueError("Share truncation algorithm 1 works only for 3 parites.")
 
         # RSPointer - public ops, Tensor Pointer - Private ops
         ptr_list = []
@@ -117,13 +117,13 @@ class ABY3(metaclass=Protocol):
         else:
             raise ValueError("{ptr_name} not supported.")
 
-        share_ptrs = ABY3.trunc1(ptr_list, x.shape, session)
+        share_ptrs = ABY3.truncation_algorithm1(ptr_list, x.shape, session)
         result = MPCTensor(shares=share_ptrs, session=session, shape=x.shape)
 
         return result
 
     @staticmethod
-    def trunc2(x: MPCTensor, session: Session) -> MPCTensor:
+    def truncation_algorithm2(x: MPCTensor, session: Session) -> MPCTensor:
         """Truncates the MPCTensor by scale factor using trunc2 algorithm.
 
         Args:
@@ -133,7 +133,7 @@ class ABY3(metaclass=Protocol):
         Returns:
             MPCTensor: truncated MPCTensor.
 
-        TODO : The trunc2 algorithm is erroneous, to be optimized.
+        TODO : The truncation algorithm 2 is erroneous, to be optimized.
         """
         r, rPrime = ABY3.get_truncation_pair(x, session)
         scale = session.config.encoder_base ** session.config.encoder_precision
@@ -166,7 +166,7 @@ class ABY3(metaclass=Protocol):
             rPrime.append(rst)
             r.append(rst)
 
-        r = ABY3.trunc1(r, x.shape, session)
+        r = ABY3.truncation_algorithm1(r, x.shape, session)
         r_mpc = MPCTensor(shares=r, session=session, shape=x.shape)
         rPrime_mpc = MPCTensor(shares=rPrime, session=session, shape=x.shape)
         return r_mpc, rPrime_mpc
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -87,6 +87,55 @@ class ABY3(metaclass=Protocol):
         return ptr_list
 
     @staticmethod
+    def truncate(x: MPCTensor, session: Session) -> MPCTensor:
+        """Perfoms the ABY3 truncation algorithm.
+
+        Args:
+            x (MPCTensor): input tensor
+            session (Session) : session of the input tensor.
+
+        Returns:
+            MPCTensor: truncated MPCTensor.
+
+        Raises:
+            ValueError : parties involved in the computation is not equal to three.
+
+        TODO :Switch to trunc2 algorithm  as it is communication efficient.
+        """
+        if session.nr_parties != 3:
+            raise ValueError("Share trunc1 algorithm works only for 3 parites.")
+
+        share_ptrs = ABY3.trunc1(x.share_ptrs, x.shape, session)
+        result = MPCTensor(shares=share_ptrs, session=session, shape=x.shape)
+
+        return result
+
+    @staticmethod
+    def trunc2(x: MPCTensor, session: Session) -> MPCTensor:
+        """Truncates the MPCTensor by scale factor using trunc2 algorithm.
+
+        Args:
+            x (MPCTensor): input tensor
+            session (Session) : session of the input tensor.
+
+        Returns:
+            MPCTensor: truncated MPCTensor.
+
+        TODO : The trunc2 algorithm is erroneous, to be optimized.
+        """
+        r, rPrime = ABY3.getTruncationPair(x, session)
+        scale = session.config.encoder_base ** session.config.encoder_precision
+        # op = getattr(operator,"sub")
+        x_rp = x - rPrime
+        x_rp = x_rp.reconstruct(decode=False) // scale
+        zero = torch.tensor([0])
+        x_rp = MPCTensor(shares=[x_rp, zero, zero], session=x.session, shape=x.shape)
+
+        result = r + x_rp
+
+        return result
+
+    @staticmethod
     def getTruncationPair(x: MPCTensor, session: Session) -> Tuple[MPCTensor]:
         """Generates truncation pair for the given MPCTensor.
 
@@ -109,36 +158,3 @@ class ABY3(metaclass=Protocol):
         r_mpc = MPCTensor(shares=r, session=session, shape=x.shape)
         rPrime_mpc = MPCTensor(shares=rPrime, session=session, shape=x.shape)
         return r_mpc, rPrime_mpc
-
-    @staticmethod
-    def truncate(x: MPCTensor, session: Session) -> MPCTensor:
-        """Truncates the MPCTensor by scale factor using trunc2 algorithm.
-
-        Args:
-            x (MPCTensor): input tensor
-            session (Session) : session of the input tensor.
-
-        Returns:
-            MPCTensor: truncated MPCTensor.
-
-        Raises:
-            ValueError : parties involved in the computation is not equal to three.
-        """
-        if session.nr_parties != 3:
-            raise ValueError("Share trunc1 algorithm works only for 3 parites.")
-
-        share_ptrs = ABY3.trunc1(x.share_ptrs, x.shape, session)
-        result = MPCTensor(shares=share_ptrs, session=session, shape=x.shape)
-
-        # TODO below trunc2 algorithm erroneous, to be optimized.
-        """r, rPrime = ABY3.getTruncationPair(x, session)
-        scale = session.config.encoder_base ** session.config.encoder_precision
-        # op = getattr(operator,"sub")
-        x_rp = x - rPrime
-        x_rp = x_rp.reconstruct(decode=False) // scale
-        zero = torch.tensor([0])
-        x_rp = MPCTensor(shares=[x_rp, zero, zero], session=x.session, shape=x.shape)
-
-        result = r + x_rp"""
-
-        return result
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -61,12 +61,12 @@ class ABY3(metaclass=Protocol):
 
     @staticmethod
     def trunc1(
-        ptr_list: List["ReplicatedSharedTensor"], shape: torch.Size, session: Session
+        ptr_list: List[torch.Tensor], shape: torch.Size, session: Session
     ) -> List["ReplicatedSharedTensor"]:
         """Performs the ABY3 trunc1 algorithm.
 
         Args:
-            ptr_list (List[ReplicatedSharedTensor]): Tensors to truncate
+            ptr_list (List[torch.Tensor]): Tensors to truncate
             shape(torch.Size) : shape of tensor values
             session(Session) : session the tensor belong to
 
@@ -79,8 +79,7 @@ class ABY3(metaclass=Protocol):
         base = session.config.encoder_base
         precision = session.config.encoder_precision
         scale = base ** precision
-        x1 = ptr_list[0].get_copy().shares[0]
-        x2, x3 = ptr_list[1].get_copy().shares
+        x1, x2, x3 = [share.get_copy() for share in ptr_list]
         x1_trunc = x1 >> precision if base == 2 else x1 // scale
         x_trunc = (x2 + x3) >> precision if base == 2 else (x2 + x3) // scale
         shares = [x1_trunc, x_trunc - rand_value, rand_value]
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -121,7 +121,11 @@ class ABY3(metaclass=Protocol):
         Returns:
             MPCTensor: truncated MPCTensor.
         """
-        r, rPrime = ABY3.getTruncationPair(x, session)
+        share_ptrs = ABY3.trunc1(x.share_ptrs, x.shape, session)
+        result = MPCTensor(shares=share_ptrs, session=session, shape=x.shape)
+
+        # TODO trunc1 algorithm erroneous, to be optimized.
+        """r, rPrime = ABY3.getTruncationPair(x, session)
         scale = session.config.encoder_base ** session.config.encoder_precision
         # op = getattr(operator,"sub")
         x_rp = x - rPrime
@@ -129,6 +133,6 @@ class ABY3(metaclass=Protocol):
         zero = torch.tensor([0])
         x_rp = MPCTensor(shares=[x_rp, zero, zero], session=x.session, shape=x.shape)
 
-        result = r + x_rp
+        result = r + x_rp"""
 
         return result
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -137,7 +137,6 @@ class ABY3(metaclass=Protocol):
         """
         r, rPrime = ABY3.get_truncation_pair(x, session)
         scale = session.config.encoder_base ** session.config.encoder_precision
-        # op = getattr(operator,"sub")
         x_rp = x - rPrime
         x_rp = x_rp.reconstruct(decode=False) // scale
         zero = torch.tensor([0])
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -62,7 +62,7 @@ class ABY3(metaclass=Protocol):
     @staticmethod
     def trunc1(
         ptr_list: List[torch.Tensor], shape: torch.Size, session: Session
-    ) -> List["ReplicatedSharedTensor"]:
+    ) -> List[ReplicatedSharedTensor]:
         """Performs the ABY3 trunc1 algorithm.
 
         Args:
@@ -79,7 +79,7 @@ class ABY3(metaclass=Protocol):
         base = session.config.encoder_base
         precision = session.config.encoder_precision
         scale = base ** precision
-        x1, x2, x3 = [share.get_copy() for share in ptr_list]
+        x1, x2, x3 = ptr_list
         x1_trunc = x1 >> precision if base == 2 else x1 // scale
         x_trunc = (x2 + x3) >> precision if base == 2 else (x2 + x3) // scale
         shares = [x1_trunc, x_trunc - rand_value, rand_value]
@@ -99,13 +99,25 @@ class ABY3(metaclass=Protocol):
 
         Raises:
             ValueError : parties involved in the computation is not equal to three.
+            ValueError : Invalid MPCTensor share pointers.
 
         TODO :Switch to trunc2 algorithm  as it is communication efficient.
         """
         if session.nr_parties != 3:
             raise ValueError("Share trunc1 algorithm works only for 3 parites.")
 
-        share_ptrs = ABY3.trunc1(x.share_ptrs, x.shape, session)
+        # RSPointer - public ops, Tensor Pointer - Private ops
+        ptr_list = []
+        ptr_name = x.share_ptrs[0].__name__
+        if ptr_name == "ReplicatedSharedTensorPointer":
+            ptr_list.append(x.share_ptrs[0].get_shares()[0].get_copy())
+            ptr_list.extend(x.share_ptrs[1].get_copy().shares)
+        elif ptr_name == "TensorPointer":
+            ptr_list = [share.get_copy() for share in x.share_ptrs]
+        else:
+            raise ValueError("{ptr_name} not supported.")
+
+        share_ptrs = ABY3.trunc1(ptr_list, x.shape, session)
         result = MPCTensor(shares=share_ptrs, session=session, shape=x.shape)
 
         return result
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -18,6 +18,8 @@ from sympc.tensor import MPCTensor
 from sympc.tensor import ReplicatedSharedTensor
 from sympc.tensor.tensor import SyMPCTensor
 
+gen = csprng.create_random_device_generator()
+
 
 class ABY3(metaclass=Protocol):
     """ABY3 Protocol Implementation."""
@@ -52,7 +54,7 @@ class ABY3(metaclass=Protocol):
         if not self.security_type == other.security_type:
             return False
 
-        if not type(self).__name__ == type(other).__name__:
+        if not type(self) == type(other):
             return False
 
         return True
@@ -71,7 +73,6 @@ class ABY3(metaclass=Protocol):
         Returns:
             List["ReplicatedSharedTensor"] : Truncated shares.
         """
-        gen = csprng.create_random_device_generator()
         rand_value = torch.empty(size=shape, dtype=session.tensor_type).random_(
             generator=gen
         )
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -148,7 +148,7 @@ class ABY3(metaclass=Protocol):
         return result
 
     @staticmethod
-    def get_truncation_pair(x: MPCTensor, session: Session) -> Tuple[MPCTensor]:
+    def _get_truncation_pair(x: MPCTensor, session: Session) -> Tuple[MPCTensor]:
         """Generates truncation pair for the given MPCTensor.
 
         Args:
--- a/src/sympc/protocol/aby3/aby3.py
+++ b/src/sympc/protocol/aby3/aby3.py
@@ -1,6 +1,6 @@
 """ABY3 Protocol.
 
-ABY3 : Mixed Protocol for Machine Learning.
+ABY3 : A Mixed Protocol Framework for Machine Learning.
 https://eprint.iacr.org/2018/403.pdf
 """
 # stdlib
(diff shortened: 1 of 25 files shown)


src/sympc/protocol/falcon/falcon.py
```diff
@@ -71,6 +73,82 @@ def __eq__(self, other: Any) -> bool:
 
         return True
 
+    @staticmethod
+    def trunc1(
```
gmuraru, line 77: Q: If this operation is specific to ABY3 could we move it to a folder named `aby3` in `protocols` folder (as we have with falcon). After that, we can call that function here by `ABY3.truncate`.

rasswanth-s: Sure George, I would change it.

src/sympc/tensor/mpc_tensor.py
```diff
@@ -780,8 +785,12 @@ def truncate(
...
             )
             result = input_tensor.truediv(scale)
-        else:
-            result = input_tensor
+        elif (
+            op_str in TRUNC_OPS
+            and (not is_private)
+            and self.session.protocol.share_class == ReplicatedSharedTensor
+        ):
+            result = ABY3.truncate(input_tensor, self.session)
```
kamathhrishi, line 793: An else statement is required.

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -338,16 +338,43 @@ def mul(self, y: Union[int, float, torch.Tensor]) -> "ReplicatedSharedTensor":
...
+            ReplicatedSharedTensor: Result of the operation.
 
         Raises:
-            NotImplementedError: Raised when implementation not present
+            ValueError: If y is not an integer or LongTensor.
         """
-        raise NotImplementedError
+        if not isinstance(y, (int, torch.LongTensor)):
+            raise ValueError("Div works (for the moment) only with integers!")
+
```
kamathhrishi, line 355: Sanity check?

kamathhrishi (changes requested):
Some comments

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -338,16 +338,43 @@ def mul(self, y: Union[int, float, torch.Tensor]) -> "ReplicatedSharedTensor":
 
         return tensor
 
-    def truediv(self, y):
+    def truediv(self, y: Union[int, float, torch.Tensor]) -> "ReplicatedSharedTensor":
         """Apply the "div" operation between "self" and "y".
 
         Args:
-            y: self/y
+            y (Union[int, float, torch.Tensor]): Denominator.
```
kamathhrishi, line 345: y should only be an int or torch.LongTensor or torch.Tensor right?

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -286,41 +286,35 @@ def test_rst_resolve_pointer(get_clients) -> None:
...
+def test_ops_public_mul_integer_matrix(get_clients, security):
+    parties = get_clients(3)
     protocol = Falcon(security)
-    session = Session(protocol=protocol, parties=parties, config=config)
+    session = Session(protocol=protocol, parties=parties)
     SessionManager.setup_mpc(session)
 
-    secret = torch.tensor([[-100, 20, 30], [-90, 1000, 1], [1032, -323, 15]])
-    value = torch.tensor([[-1, 2, 3], [-9, 10, 1], [32, -23, 5]])
+    secret = torch.Tensor([[0.125, -1.25], [-4.25, 4]])
```
kamathhrishi, line 314: Could you have a mixed value? Large integer values and with floating-point. As well as small values close to and lesser than 0.

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,144 @@
...
+        x1 = ptr_list[0].get_copy().shares[0]
+        x2, x3 = ptr_list[1].get_copy().shares
+        x1_trunc = x1 >> precision if base == 2 else x1 // scale
+        x_trunc = (x2 + x3) >> precision if base == 2 else (x2 + x3) // scale
+        shares = [x1_trunc, x_trunc - rand_value, rand_value]
+        ptr_list = ReplicatedSharedTensor.distribute_shares(shares, session)
+        return ptr_list
+
+    @staticmethod
+    def getTruncationPair(x: MPCTensor, session: Session) -> Tuple[MPCTensor]:
```
kamathhrishi, line 90: Just for readability sake, I would suggest moving this function to the end of the file so that we could come back later to it.

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,144 @@
...
+        Raises:
+            ValueError : parties involved in the computation is not equal to three.
+        """
+        if session.nr_parties != 3:
+            raise ValueError("Share trunc1 algorithm works only for 3 parites.")
+
+        share_ptrs = ABY3.trunc1(x.share_ptrs, x.shape, session)
+        result = MPCTensor(shares=share_ptrs, session=session, shape=x.shape)
+
+        # TODO below trunc2 algorithm erroneous, to be optimized.
```
kamathhrishi, line 133: You can keep this truncation algorithm as a function at the end of the file with a note that implementation isn't accurate.

src/sympc/tensor/replicatedshare_tensor.py
rasswanth-s: Yeah , an error is raised when it is not int or tensor, in future when we add support for float,we could remove the sanity check.

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,160 @@
...
+        Args:
+            other (Any): Object to compare
+
+        Returns:
+            bool: True if equal False if not.
+        """
+        if not self.security_type == other.security_type:
+            return False
+
+        if not type(self).__name__ == type(other).__name__:
```
gmuraru, line 55: Q: Can't we simple check if ```type(self) != type(other)```?

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Args:
+            ptr_list (List[ReplicatedSharedTensor]): Tensors to truncate
+            shape(torch.Size) : shape of tensor values
+            session(Session) : session the tensor belong to
+
+        Returns:
+            List["ReplicatedSharedTensor"] : Truncated shares.
+        """
+        gen = csprng.create_random_device_generator()
```
gmuraru, line 74: MAYBE: This generator should stay I think in the init. We create a generator and then we use the same generator for subsequent calls.

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,160 @@
...
+        if not self.security_type == other.security_type:
+            return False
+
+        if not type(self).__name__ == type(other).__name__:
+            return False
+
+        return True
+
+    @staticmethod
+    def trunc1(
```
gmuraru, line 61: Q: Could you add some tests for this? (to test the protocol not the operation from RST)

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Sure , I will set it as global variable like we do beaver protocol.

tests/sympc/protocol/aby3/aby3_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+def test_invalid_parties_trunc(get_clients) -> None:
+    parties = get_clients(2)
+    session = Session(parties=parties)
+
+    with pytest.raises(ValueError):
+        ABY3.truncate(None, session)
+
+
+def test_trunc1(get_clients) -> None:
```
gmuraru, line 38: MAYBE: A test where base is 2?

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,161 @@
...
+    def __eq__(self, other: Any) -> bool:
+        """Check if "self" is equal with another object given a set of attributes to compare.
+
+        Args:
+            other (Any): Object to compare
+
+        Returns:
+            bool: True if equal False if not.
+        """
+        if not self.security_type == other.security_type:
```
gmuraru, line 54: MAYBE: Add `self.security_type != other.security_type`

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,161 @@
...
+            return False
+
+        if not type(self) == type(other):
```
gmuraru, line 57: Also here, use `!=`

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,161 @@
...
+        x_rp = x_rp.reconstruct(decode=False) // scale
+        zero = torch.tensor([0])
+        x_rp = MPCTensor(shares=[x_rp, zero, zero], session=x.session, shape=x.shape)
+
+        result = r + x_rp
+
+        return result
+
+    @staticmethod
+    def getTruncationPair(x: MPCTensor, session: Session) -> Tuple[MPCTensor]:
```
gmuraru, line 140: OUGHTTO: Use ```snake style``` for methods name.
```get_truncation_pair``` since that is how we use it in our codebase.

File: src/sympc/protocol/aby3/aby3.py [hunk 15]
```diff
@@ -0,0 +1,161 @@
...
+        if session.nr_parties != 3:
+            raise ValueError("Share trunc1 algorithm works only for 3 parites.")
+
+        share_ptrs = ABY3.trunc1(x.share_ptrs, x.shape, session)
+        result = MPCTensor(shares=share_ptrs, session=session, shape=x.shape)
+
+        return result
+
+    @staticmethod
+    def trunc2(x: MPCTensor, session: Session) -> MPCTensor:
```
gmuraru, line 115: Q: Could we find a better name for ```trunc1```/```trunc2```?

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,161 @@
...
+        if not self.security_type == other.security_type:
+            return False
+
+        if not type(self) == type(other):
+            return False
+
+        return True
+
+    @staticmethod
+    def trunc1(
```
gmuraru, line 63: Q: Shouldn't we delete it from here? (if we have it in ABY3?)

src/sympc/protocol/falcon/falcon.py
```diff
@@ -66,7 +66,7 @@ def __eq__(self, other: Any) -> bool:
         if not self.security_type == other.security_type:
```
gmuraru, line 66: Q: Could you also change it here. To use `!=`

src/sympc/protocol/falcon/falcon.py
```diff
@@ -66,7 +66,7 @@ def __eq__(self, other: Any) -> bool:
...
             return False
 
-        if not type(self).__name__ == type(other).__name__:
+        if not type(self) == type(other):
```
gmuraru, line 69: Also here

src/sympc/protocol/falcon/falcon.py
```diff
@@ -92,6 +92,8 @@ def mul_master(x: MPCTensor, y: MPCTensor, session: Session) -> MPCTensor:
             NotImplementedError: Raised when implementation not present
 
         """
+        from sympc.protocol import ABY3
```
gmuraru, line 95: Q: Can't we import this at the top of the module? Or we get circular import?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -117,6 +119,8 @@ def mul_master(x: MPCTensor, y: MPCTensor, session: Session) -> MPCTensor:
                 z_shares, x.session
             )
             result = MPCTensor(shares=reshared_shares, session=x.session)
+            result.shape = MPCTensor._get_shape("mul", x.shape, y.shape)  # for prrs
```
gmuraru, line 122: MAYBE: You reassign the result on the next line. I think this line would have no effect.

src/sympc/protocol/fss/fss.py
```diff
@@ -252,7 +252,7 @@ def __eq__(self, other: Any) -> bool:
         if not self.security_type == other.security_type:
             return False
 
-        if not type(self).__name__ == type(other).__name__:
+        if not type(self) == type(other):
```
gmuraru, line 255: MAYBE: Also here (to use `!=`)

src/sympc/tensor/mpc_tensor.py
```diff
@@ -769,8 +770,12 @@ def truncate(
             result (MPCTensor): Truncated result
         """
         # Truncation should be added for RSTensor
+        from sympc.protocol import ABY3
```
gmuraru, line 773: Q: This is the same case: if you move them at the top of the module they don't work?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -338,16 +338,43 @@ def mul(self, y: Union[int, float, torch.Tensor]) -> "ReplicatedSharedTensor":
...
+        Returns:
+            ReplicatedSharedTensor: Result of the operation.
 
         Raises:
-            NotImplementedError: Raised when implementation not present
+            ValueError: If y is not an integer or LongTensor.
         """
-        raise NotImplementedError
+        if not isinstance(y, (int, torch.LongTensor)):
+            raise ValueError("Div works (for the moment) only with integers!")
```
gmuraru, line 354: MAYBE: Also Long tensors

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Yeah , george, I kept it only to aby3.

src/sympc/tensor/mpc_tensor.py
rasswanth-s: Yeah,we get a circular import error.

src/sympc/protocol/aby3/aby3.py
rasswanth-s: I followed the convention from the ABY3 paper,so it would be easy to refer to, any suggestions for a better name?

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Shape is required for prrs generation, when I pass result to `ABY3.truncate`,it infers it from result.shape
https://github.com/OpenMined/SyMPC/blob/f38d35bb6d52e6453948c9e25896bd7ae81da1db/src/sympc/protocol/falcon/falcon.py#L122

tests/sympc/protocol/aby3/aby3_test.py
rasswanth-s: Yeah , tests where base is not 2 right.Good catch,I forgot to add  that scenario.

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -286,41 +286,35 @@ def test_rst_resolve_pointer(get_clients) -> None:
     assert share_pt_name == "ReplicatedSharedTensorPointer"
 
 
-@pytest.mark.parametrize("parties", [3, 5, 7])
-@pytest.mark.parametrize("security", ["malicious", "semi-honest"])
-def test_ops_public_mul_integer(get_clients, parties, security):
-    # Not encoding because truncation hasn't been implemented yet for Falcon
-    config = Config(encoder_base=1, encoder_precision=0)
+@pytest.mark.parametrize("security", ["semi-honest"])  # malicious to be added
+def test_ops_public_mul_integer(get_clients, security):
```
kamathhrishi, line 290: I think we should still support public integer multiplication for n parties. Could you also add a test for it?

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+        base = session.config.encoder_base
+        precision = session.config.encoder_precision
+        scale = base ** precision
+        x1 = ptr_list[0].get_copy().shares[0]
+        x2, x3 = ptr_list[1].get_copy().shares
+        x1_trunc = x1 >> precision if base == 2 else x1 // scale
+        x_trunc = (x2 + x3) >> precision if base == 2 else (x2 + x3) // scale
+        shares = [x1_trunc, x_trunc - rand_value, rand_value]
+        ptr_list = ReplicatedSharedTensor.distribute_shares(shares, session)
```
kamathhrishi, line 87: You could keep this. But, what we could do is remove resharing in multiplication. Truncation requires 3 values. We could use z1,z2&z3 after Falcon mult for truncation and then perform 2-3 resharing rather than do it twice?

tests/sympc/tensor/replicatedshare_tensor_test.py
rasswanth-s: The truncation algorithm 1 works only  for 3 three parties.

kamathhrishi: Exactly why I am saying we could support public integer multiplication. Before truncation, I was doing it for n parties. Similar to that.

kamathhrishi: We don't need truncation when base is 1 and precision is 0.

rasswanth-s: Yeah , without truncation it would work, I would add it.

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Yeah thought would be communication efficient, Sure I would change it.

src/sympc/tensor/replicatedshare_tensor.py
kamathhrishi: No I meant the type annotation.

src/sympc/tensor/mpc_tensor.py
```diff
@@ -769,8 +770,12 @@ def truncate(
...
+        from sympc.tensor import ReplicatedSharedTensor
+
+        result = None
         if (
-            op_str in {"mul", "matmul", "conv2d", "conv_transpose2d"}
+            op_str in TRUNC_OPS
```
kamathhrishi, line 778: Truncated_ops sounds better?

src/sympc/protocol/aby3/aby3.py
rasswanth-s: We do share generation at distribution ,so I implemented in this manner.

src/sympc/protocol/aby3/aby3.py
kamathhrishi: truncation_algorithm1 and truncation_algorithm2 ?

src/sympc/tensor/mpc_tensor.py
rasswanth-s: Yeah okay for me. :+1:

src/sympc/protocol/aby3/aby3.py
gmuraru: Yep, let's go with this

src/sympc/protocol/falcon/falcon.py
gmuraru: Ahh..you are right. I missed where you use result as an argument.

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        rand_value = torch.empty(size=shape, dtype=session.tensor_type).random_(
+            generator=gen
+        )
+        base = session.config.encoder_base
+        precision = session.config.encoder_precision
+        scale = base ** precision
+        x1, x2, x3 = ptr_list
+        x1_trunc = x1 >> precision if base == 2 else x1 // scale
+        x_trunc = (x2 + x3) >> precision if base == 2 else (x2 + x3) // scale
```
gmuraru, line 84: Q: if we have ```base == 2``` doesn't this mean `x2` and `x3` are only 0 or 1?

rasswanth-s: No George , They are arithmetic values only, The reason I did was ,In fixed point , generally we use base 2 for computational reasons, for scaling down instead of dividing by `2**d` , an arithmetic right shift is performed for O(1) computation.

gmuraru: Ahhh..yep, you are right. I was thinking at `ring_size` of size 2

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,172 @@
...
+        if self.security_type != other.security_type:
+            return False
+
+        if type(self) != type(other):
+            return False
+
+        return True
+
+    @staticmethod
+    def truncation_algorithm1(
```
gmuraru, line 63: Q: Can't we use parallel execution for this? (and in that case, we won't have to locally get the shares)

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,172 @@
...
+import torch
+import torchcsprng as csprng
+
+from sympc.protocol.protocol import Protocol
+from sympc.session import Session
+from sympc.tensor import MPCTensor
+from sympc.tensor import ReplicatedSharedTensor
+from sympc.tensor.tensor import SyMPCTensor
+
+gen = csprng.create_random_device_generator()
```
gmuraru, line 21: Q: From the protocol I see that there is a generator shared only between party 2 and 3

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,172 @@
...
+        if session.nr_parties != 3:
+            raise ValueError("Share truncation algorithm 1 works only for 3 parites.")
+
+        # RSPointer - public ops, Tensor Pointer - Private ops
+        ptr_list = []
+        ptr_name = x.share_ptrs[0].__name__
+        if ptr_name == "ReplicatedSharedTensorPointer":
+            ptr_list.append(x.share_ptrs[0].get_shares()[0].get_copy())
+            ptr_list.extend(x.share_ptrs[1].get_copy().shares)
+        elif ptr_name == "TensorPointer":
```
gmuraru, line 115: Q: Can we reach this case? (won't we always have ```ReplcatedShareTensorPointers```?

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,172 @@
...
+        x_rp = x_rp.reconstruct(decode=False) // scale
+        zero = torch.tensor([0])
+        x_rp = MPCTensor(shares=[x_rp, zero, zero], session=x.session, shape=x.shape)
+
+        result = r + x_rp
+
+        return result
+
+    @staticmethod
+    def get_truncation_pair(x: MPCTensor, session: Session) -> Tuple[MPCTensor]:
```
gmuraru, line 151: Q: Can we name this ```_get_truncation_pair``` (with leading `_`) only to show the "dev" that this is intended to be used only internally in this file?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -110,13 +111,10 @@ def mul_master(x: MPCTensor, y: MPCTensor, session: Session) -> MPCTensor:
...
             )(args)
 
-            z_shares = [share.get() for share in z_shares_ptrs]
-
-            # Convert 3-3 shares to 2-3 shares by resharing
-            reshared_shares = ReplicatedSharedTensor.distribute_shares(
-                z_shares, x.session
-            )
-            result = MPCTensor(shares=reshared_shares, session=x.session)
+            print(z_shares_ptrs)
```
gmuraru, line 114: MAYBE: We should remove this, right?

src/sympc/protocol/fss/fss.py
```diff
@@ -249,10 +249,10 @@ def __eq__(self, other: Any) -> bool:
         Returns:
             bool: True if equal False if not.
         """
-        if not self.security_type == other.security_type:
+        if self.security_type != other.security_type:
             return False
 
-        if not type(self).__name__ == type(other).__name__:
+        if type(self) != type(other):
```
gmuraru, line 255: Thank you :+1:

src/sympc/tensor/mpc_tensor.py
gmuraru: Don't forget about this :smile:

gmuraru (commented):
I left some comments. I am also ok with merging this as it is now and make some of the fixes (like the one with `parallel_execution` and shared `generator` between party 2 and 3 in another PR.

src/sympc/protocol/aby3/aby3.py
rasswanth-s: I wanted to do parallel execution at start, in that scenario parties mutually communicate with each other ,we require a P2P for that,share truncation algorithm 1 involves increased round complexity ,when we execute it without, I highly optimized it for our orchestrator model, so we could have only one round complexity. Eventually when we move to truncation `2` ,we delegate it locally using parallel execution.

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Yeah, in the p2p model, parties local generate random values using common prrs, and reveal only to parties 2 & 3 for truncation algorithm1, when we take this model to orchestrator , I found we had several round  complexity for `trunc1`,I made it a common generator, but when we switch truncation algorithm 2  ,it would be optimized.

src/sympc/protocol/aby3/aby3.py
rasswanth-s: During `Falcon_multiplication protocol` for private ops we get `Tensor Pointer`,       
https://github.com/OpenMined/SyMPC/blob/99c559bfd82ee9c2f61707fbf258f83f5bad0223/src/sympc/protocol/falcon/falcon.py#L151
We delayed the resharing to reduced round complexity

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Sure,I would change it :+1:

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -286,30 +286,48 @@ def test_rst_resolve_pointer(get_clients) -> None:
...
+    assert np.allclose(result.reconstruct(), expected_res, atol=1e-3)
 
 
-@pytest.mark.parametrize("parties", [3, 5, 7])
-@pytest.mark.parametrize("security", ["malicious", "semi-honest"])
-def test_ops_public_mul_integer_matrix(get_clients, parties, security):
-    # Not encoding because truncation hasn't been implemented yet for Falcon
+@pytest.mark.parametrize("base, precision", [(2, 16), (2, 17), (10, 3), (10, 4)])
+@pytest.mark.parametrize("security", ["semi-honest"])  # malicous to be addded
+def test_ops_public_mul_integer_matrix(get_clients, security, base, precision):
```
kamathhrishi, line 311: Could you change the name? The integer shouldn't be here.

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -286,30 +286,48 @@ def test_rst_resolve_pointer(get_clients) -> None:
...
 
 
-@pytest.mark.parametrize("parties", [3, 5, 7])
-@pytest.mark.parametrize("security", ["malicious", "semi-honest"])
-def test_ops_public_mul_integer(get_clients, parties, security):
-    # Not encoding because truncation hasn't been implemented yet for Falcon
-    config = Config(encoder_base=1, encoder_precision=0)
+@pytest.mark.parametrize("base, precision", [(2, 16), (2, 17), (10, 3), (10, 4)])
+@pytest.mark.parametrize("security", ["semi-honest"])  # malicious to be added
+def test_ops_public_mul_integer(get_clients, security, base, precision):
```
kamathhrishi, line 291: Name change

kamathhrishi (approved):
Looks good to me. Just rename two of the unit tests and then you can merge.

src/sympc/protocol/aby3/aby3.py
```diff
@@ -0,0 +1,172 @@
...
+            session (Session) : session of the input tensor.
+
+        Returns:
+            MPCTensor: truncated MPCTensor.
+
+        TODO : The truncation algorithm 2 is erroneous, to be optimized.
+        """
+        r, rPrime = ABY3.get_truncation_pair(x, session)
+        scale = session.config.encoder_base ** session.config.encoder_precision
+        # op = getattr(operator,"sub")
```
gmuraru, line 140: MAYBE: Remove this

tests/sympc/tensor/replicatedshare_tensor_test.py
```diff
@@ -318,9 +336,39 @@ def test_ops_public_mul_integer_matrix(get_clients, parties, security):
...
+    with pytest.raises(ValueError):
+        rst >> 1.55
+
+
+def test_rshift() -> None:
+    secret = 10.25
+    rst = ReplicatedSharedTensor(shares=[secret])
+    rst = rst >> 2
+    expected_res = rst.fp_encoder.encode(secret) >> 2
+    assert rst.shares[0] == expected_res
```
gmuraru, line 374: Q: Shouldn't we test the scenario where we do reconstruction and the values are "approximately" the same? (also for the `truediv`)?

gmuraru (approved):
In rest LGTM! :rocket:

tests/sympc/tensor/replicatedshare_tensor_test.py
rasswanth-s: Sure,I will do it :+1:

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Discussed in slack.

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Discussed in slack.

tests/sympc/tensor/replicatedshare_tensor_test.py
rasswanth-s: For the MPCTensor it would not work , as we would get the carry bit issue.

rasswanth-s: Dicussed in slack.

PR #262, opened 2021-06-27
Fix the crypto provider generation for FSS

LaRiffle:
## Description
For FSS key generation, the `primitive` is not a tuple but a tensor. Hence, the `itertools.chain(*primitive)` iterates _through_ the tensor which doesn't make sense (but didn't fail for some reason). So I convert it to a tuple!

@gmuraru perhaps point this PR to one of your branch? Only one commit is mine here


gmuraru (approved):
LGTM!

Issue #160, opened 2021-05-09
Replicated Sharing Tensor (RST): Matmul Operation


kamathhrishi:
## Feature Description
Implement operation to RSS Tensor for the matmul of private and public value.
RSS is described in #145
Tasklist
- [x] Semi Honest
- [x] Malicious

PR #264, opened 2021-06-29
Falcon  Multiplication(M) , Matrix Multiplication(S,M) , Beaver Triple Generation.(S-Semi-honest,M-Malicious)

### LINKED ISSUE #158: Replicated Sharing Tensor (RST): Mul Operation

rasswanth-s:
## Description
Implementation of Falcon Malicious Multiplication ,Matrix Multiplication(Semi-honest,Malicious), fixes #158 , #160, #172.

Note :
I have implemented as per the falcon protocol, there are some cases , where round complexity could be reduced due to our orchestrator model, feel free to point out.

## How has this been tested?
Added tests for the same.

## Checklist


tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -108,17 +112,21 @@ def test_mul_private_exception_nothreeparties(get_clients, parties):
         tensor1 * tensor2
```
kamathhrishi, line 113: I was wondering if there was any way you could test if an exception was being raised in case of malicious share values.

src/sympc/protocol/beaver/beaver.py
```diff
@@ -49,42 +52,71 @@ def _get_triples(
...
+            nr_parties=nr_parties,
+            tensor_type=torch.long,
+            config=config,
+        )
+
+        c_val = cmd(a_rand, b_rand, **kwargs)
+        c_shares = MPCTensor.generate_shares(
+            secret=c_val, nr_parties=nr_parties, tensor_type=torch.long, config=config
+        )
+    elif session.protocol.share_class == ReplicatedSharedTensor:
```
kamathhrishi, line 98: Just add an else statement raising the Invalid share class error.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
         Args:
             x (MPCTensor): Secret
             y (MPCTensor): Another secret
             session (Session): Session the tensors belong to
+            op_str (str): Operation string.
+            kwargs_ (dict): Kwargs for some operations like conv2d
 
         Returns:
-            shares (ReplicatedSharedTensor): Shares in terms of ReplicatedSharedTensor.
+            MPCTensor: Result of the operation.
```
kamathhrishi, line 97: NIT: Add the return variable name and annotation.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
+        x: MPCTensor,
+        y: MPCTensor,
+        session: Session,
+        op_str: str,
+        kwargs_: Dict[Any, Any],
+        truncate: bool = True,
+    ) -> MPCTensor:
+        """Falcon semihonest multiplication.
+
+        Performs Falcon's mul implementation, gets and reshares mul results and distributes shares.
```
kamathhrishi, line 130: _Performs Falcon's mul implementation, gets and reshares mul results and distributes shares_ , Change it to _Performs Falcon's mul implementation, add masks and performs resharing_ 

The description I had put earlier was verbose

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
+    ) -> Tuple[ReplicatedSharedTensor, ReplicatedSharedTensor]:
+        """Falcon mask.
+
+        Args:
+            x_sh (ReplicatedSharedTensor): X share
+            y_sh (ReplicatedSharedTensor) : Y share
+            op_str (str): Operator
+
+        Returns:
+            Tuple[ReplicatedSharedTensor,ReplicatedSharedTensor] : Masked values.
```
kamathhrishi, line 238: The format should be same as args.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
+
+        Args:
+            x (MPCTensor): Secret
+            y (MPCTensor): Another secret
+            session (Session): Session the tensors belong to
+            op_str (str): Operation string.
+            kwargs_ (dict): Kwargs for some operations like conv2d
+
+        Returns:
+            MPCTensor: Result of the operation.
```
kamathhrishi, line 270: Here too

tests/sympc/protocol/falcon/falcon_test.py
rasswanth-s: I thought of the scenario, a place where we could make it raise is, modify  the triples after generation initially, this would raise the error, it has to be done locally by the parties, I tried to modify crypto store values after generation, was not able to do so,not sure if orchestrator could modify the values of the parties? @gmuraru ,is there any way for orchestrator to modify crypto store values.
Or there might some better way to raise that error, any suggestions?

kamathhrishi: I haven't been able to think of any way yet.

src/sympc/protocol/falcon/falcon.py
rasswanth-s: There was no explicit return variable name, so I followed  the convention in spdz.Shall I assign a variable and modify or leave it like this?

kamathhrishi: You can just put it like: _values (Tuple[ReplicatedSharedTensor,ReplicatedSharedTensor]): masked_values_  Don't have to create a variable for it specifically.

kamathhrishi:
The functionality looks correct, but would still require a unit test. One way I thought was to add a function in unit tests that is exactly the same as mul_malicious(https://github.com/rasswanth-s/SyMPC/blob/58a608d54eac68724482776c81dd90dbee1c8a88/src/sympc/protocol/falcon/falcon.py#L251) but adds some random value to the result of multiplication. So this does not end up testing mul_malicious as a whole but ends up testing functionality of triple_verification and falcon_mask indirectly. 

It would be great to have a way in which the orchestrator could directly manipulate values such that we test the entire mul_malicious.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,293 @@ def __eq__(self, other: Any) -> bool:
...
+            op_str (str): Operation string.
+            kwargs_ (dict): Kwargs for some operations like conv2d
+            truncate (bool) : Applies truncation on the result if set.
+
+        Returns:
+            MPCTensor: Result of the operation.
+        """
+        result = None
+        args = []
+        for index in range(0, 3):
```
gmuraru, line 145: NIT:
```
args = [(x_share, y_xshare, op_str) for x_share in y_share in zip(x.share_ptrs, y.share_ptrs)]
```

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,293 @@ def __eq__(self, other: Any) -> bool:
...
+
+        return x_sh - a_sh, y_sh - b_sh
+
+    @staticmethod
+    def mul_malicious(
+        x: MPCTensor,
+        y: MPCTensor,
+        session: Session,
+        op_str: str,
+        kwargs_: Dict[Any, Any],
```
gmuraru, line 256: MAYBE: kwargs is usually ```**kwargs``` why we need it like this?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,293 @@ def __eq__(self, other: Any) -> bool:
...
+        przs_mask = session.przs_generate_random_share(shape=shape)
         # Add PRZS Mask to z  value
-        share = z_value.get_shares()[0] + przs_mask.get_shares()[0]
+        share = z_value + przs_mask.get_shares()[0]
         return share
 
     @staticmethod
     def multiplication_protocol(
-        x: ReplicatedSharedTensor, y: ReplicatedSharedTensor
+        x: ReplicatedSharedTensor, y: ReplicatedSharedTensor, op_str: str, **kwargs
```
gmuraru, line 349: MAYBE: Type annotation for kwargs

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,293 @@ def __eq__(self, other: Any) -> bool:
...
+        result.shape = MPCTensor._get_shape(op_str, x.shape, y.shape)  # for prrs
         return result
 
+    @staticmethod
+    def triple_verification(
+        z_sh: ReplicatedSharedTensor,
+        eps: torch.Tensor,
+        delta: torch.Tensor,
+        op_str: str,
+        **kwargs,
```
gmuraru, line 176: NIT: Type annotation

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,293 @@ def __eq__(self, other: Any) -> bool:
...
+
+        return result
+
+    @staticmethod
+    def mul_semi_honest(
+        x: MPCTensor,
+        y: MPCTensor,
+        session: Session,
+        op_str: str,
+        kwargs_: Dict[Any, Any],
```
gmuraru, line 125: Q: This should be **kwargs, right?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,93 +77,293 @@ def __eq__(self, other: Any) -> bool:
...
+        primitives = crypto_store.get_primitives_from_store(
+            f"beaver_{op_str}", x_sh.shape, y_sh.shape, remove=False
+        )
+
+        a_sh, b_sh, _ = primitives
+
+        return x_sh - a_sh, y_sh - b_sh
+
+    @staticmethod
+    def mul_malicious(
```
gmuraru, line 251: Q: The implementation does not look like what they say in the paper? (in the paper, for malicious setup, they say that x_i is sent to P_i+1 and x_i+1 to P_i-1,

kamathhrishi: That's for reconstruction. Here the implementation of malicious mult is from ABY3. The reason why the reconstruction verification does not work is that parties have only a single share after computing Z, reconstruction verification requires 2 per party. So in order to still provide malicious security, ABY3 paper (actually Lindell but ABY3 extends it to arithmetic), there is this protocol where they compute beaver triplets on original shares used for mult.

kamathhrishi: It's explained in ABY3 paper section 9.1 or 9.2 clearly.

src/sympc/protocol/falcon/falcon.py
rasswanth-s: @kamathhrishi  , had this exact change in his refactoring PR, https://github.com/OpenMined/SyMPC/pull/255 , so I thought not to do here,Shall I change here or we will address in the refactor PR.

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Adding on to what Hrishi said, reconstruction detects malicious behavior ,but it also reveals the underlying secret, we do not want to do this , in the middle of the computation(leaking intermediate values unless we intend to),for verification we do the triple check, george you think of it similar like spdz , where you use beaver triples for multiplication itself, here we have a custom protocol for multiplication , but use the triple as a verifier ,to veriy that the above multiplication is correct.

src/sympc/protocol/falcon/falcon.py
kamathhrishi: I have addressed it so it will change after I merge, so you can keep it like that.

rasswanth-s: I have modified it after the merge.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,87 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
+            y (MPCTensor): Another secret
+            session (Session): Session the tensors belong to
+            op_str (str): Operation string.
+            truncate (bool) : Applies truncation on the result if set.
+            kwargs_ (Dict[Any, Any]): Kwargs for some operations like conv2d
+
+        Returns:
+            MPCTensor: Result of the operation.
+        """
+        result = None
```
gmuraru, line 143: MAYBE: Remove this

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,87 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
+
+        args = [
```
gmuraru, line 145: MAYBE: `args = list(zip(x.share_ptrs, y.share_ptrs))` (I am not sure if it is more compact and clear?)

src/sympc/protocol/falcon/falcon.py
gmuraru: @kamathhrishi yep - my bad...checking it now with the ABY3 details

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,87 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
+            for sh in zip(result.share_ptrs)
+        ]
+
+        triple_shares = parallel_execution(Falcon.triple_verification, session.parties)(
+            args, kwargs_
+        )
+
+        triple = MPCTensor(shares=triple_shares, session=x.session)
+
+        if (triple.reconstruct(decode=False) == result.reconstruct(decode=False)).all():
```
gmuraru, line 314: Q: Wouldn't be better at this step if we don't reconstruct the result of the multiplication. but reconstruct this
image

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,87 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
+            # Convert 3-3 shares to 2-3 shares by resharing
+            reshared_shares = ReplicatedSharedTensor.distribute_shares(
+                z_shares, x.session
+            )
+            result = MPCTensor(shares=reshared_shares, session=x.session)
+        result.shape = MPCTensor._get_shape(op_str, x.shape, y.shape)  # for prrs
         return result
 
+    @staticmethod
+    def triple_verification(
```
gmuraru, line 168: Q: Shouldn't we do hear something like this?
image
And in the `malicious_multiplication` reconstruct this value and if it is ```0 (zero)``` we know that the result (formed from the shares from "z") is correct.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -73,87 +77,295 @@ def __eq__(self, other: Any) -> bool:
...
         Returns:
             shares (ReplicatedSharedTensor): results in terms of ReplicatedSharedTensor.
         """
+        op = getattr(operator, op_str)
+
         z_value = (
-            x.shares[0] * y.shares[0]
-            + x.shares[1] * y.shares[0]
-            + x.shares[0] * y.shares[1]
+            op(x.shares[0], y.shares[0], **kwargs)
```
gmuraru, line 367: :+1:

src/sympc/session/session.py
```diff
@@ -278,7 +278,7 @@ def prrs_generate_random_share(
             share = ReplicatedSharedTensor(
                 shares=[share1, share2],
                 session_uuid=self.uuid,
-                config=Config(encoder_precision=0),
+                # config=Config(encoder_precision=0),
```
gmuraru, line 281: Q: Remove this?

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -318,17 +320,58 @@ def mul(self, y: Union[int, float, torch.Tensor]) -> "ReplicatedSharedTensor":
...
 
-                result = [Falcon.multiplication_protocol(self, y_tensor)]
+                result = [Falcon.multiplication_protocol(self, y_tensor, op_str)]
             else:
                 raise ValueError(
                     "Private mult between ReplicatedSharedTensors is allowed only for 3 parties"
                 )
         else:
-            result = [share * y_tensor.shares[0] for share in self.shares]
+            result = [op(share, y_tensor.shares[0]) for share in self.shares]
```
gmuraru, line 335: MAYBE: Why not directly use ```operator.mul(share, y_tensor.share[0])``` here? (and remove the `op` from above?)

gmuraru (commented):
Left some comments, I think it looks great. Nice job!

src/sympc/protocol/falcon/falcon.py
rasswanth-s: I included the op_str like you suggested,
<code>
args = [
            [x_share, y_share, op_str]
            for x_share, y_share in zip(x.share_ptrs, y.share_ptrs)
        ]
</code>

src/sympc/protocol/falcon/falcon.py
rasswanth-s: George for that part, semi -honest multiplication induces additional randomness by `PRZS` , this also is for beaver triple calculation, `PRZS` randomizes it.So we require the `reconstruction` step.

I considered several scenario's while working on it, there are scenario's where `PRZS` is required (just only semi-honest mult and without truncation) and `PRZS` is  not required when(semi-honest mult without truncation) , the same for malicious multiplication.

There are lot of optimizations which could be done,I thought to do it for another PR, after completion of basic functionalities, as the requirements might change then.

src/sympc/session/session.py
rasswanth-s: George I have reverted to use no encoding.

tests/sympc/protocol/falcon/falcon_test.py
rasswanth-s: I have added a test for the same.

kamathhrishi (approved):
Looks good to me.

src/sympc/protocol/falcon/falcon.py
rasswanth-s: I modified the reconstruction according to ABY3.

gmuraru (approved):
LGTM! :rocket:

PR #245, opened 2021-06-20
[WIP] Add rust POC

gmuraru:
## Description
Move some of the functionalities to a rust folder.

## Affected Dependencies
List any dependencies that are required for this change.

## How has this been tested?
- The tests should pass as they are now - since that functionality is tested in Python

## Checklist
- [x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [x] My changes are covered by tests

PR #268, opened 2021-07-06
[WIP] Rust POC with `Maturin`

arturomf94:
## Description

This PR is to test out the integration of a `Rust` _"backend"_ for `SyMPC` (following work from #245), with [`Maturin`](https://github.com/PyO3/maturin) as its binding framework. 

_Update: So far, this PR has got the most basic set-up, following the structure in the [`feature/rust`](https://github.com/OpenMined/PySyft/tree/feature/rust) branch from `PySyft`._ 

Some things to do: 

- [ ] Test the binding of some basic Python functionality. We already have a full project, [`sycret`](https://github.com/OpenMined/sycret) that does this, so we could probably recycle some stuff from there.
- [ ] Write Rust tests. I think there are no Rust tests so far (05-07-21), but this is definitely something that we should have so that we can integrate the code seamlessly.

## Affected Dependencies

Dependencies will be affected but this is not yet clear, so will update this section later.

## How has this been tested?
- Tests are running fine so far, and we can successfully run `maturin develop` & `cargo test` so everything's good up 'til this point. 

## Checklist


arturomf94:
@madhavajay @gmuraru - in terms of integration, what do you think would be best: to start adding part of the code-base bit by bit (different PRs) or to have one PR with all of the Rust "backend"? This is more of a long-term thing I guess but it would be nice to foresee that.

tholop:
Nice! One small thing: I feel that the structure of this kind of Maturin project can be a bit confusing. 

Here, it seems that we will have `SyMPC/src/sympc/` for the Python code, and `SyMPC/src/src/` for the Rust code. That's odd. Also, The Rust tests are in the main lib for this PR, but we might want to put them in their own directory later (the Sycret repo looks quite weird with `sycret/test/` for the Python tests and `sycret/tests/` for the Rust tests).

One suggestion would be to do something like: `SyMPC/src/python/` and `SyMPC/src/rust/` for the Python/Rust source code, and do the same thing for the tests with `SyMPC/tests/python/`, `SyMPC/tests/notebooks/` and `SyMPC/tests/rust/`.

arturomf94:
> Here, it seems that we will have SyMPC/src/sympc/ for the Python code, and SyMPC/src/src/ for the Rust code.

Indeed, I left it like that because it's the same structure as the one in `PySyft`. Although, I agree that it can be a bit confusing. Interested in reading what @gmuraru and @madhavajay think about that.

> The Rust tests are in the main lib for this PR, but we might want to put them in their own directory later

Yes! Absolutely. Agree 💯 - I left it like that because it was the default when running `cargo init --lib` but I plan to change it in the near future.

Thanks for these comments @tholop!

arturomf94:
@tholop  - I've moved the Rust tests to a separate folder. Depending on what the final structure, this folder might stay there or be at the root of the project, along with the Python tests.

gmuraru:
Could we also have a basic function added and called in the Python code?
I am thinking at the `ispointer` function.

arturomf94:
> Could we also have a basic function added and called in the Python code?
> I am thinking at the `ispointer` function.

Yes - I am thinking of adding functionality to "prove" that this works. I think is `is_pointer` might be a good function to test. I'll try with that. However, I guess the real advantage of having a Rust backend would be to make computation more efficient, so it would be good to discuss **as a team** whether we will, indeed, want to port some of the code-base to Rust and, if so, which parts.

PR #269, opened 2021-07-06
Extension of ReplicatedSharedTensor to Binary,Prime order rings.

rasswanth-s:
## Description
Extended RST to binary, prime order ring_sizes ,as it is required for higher level functionalities of Falcon.

## Affected Dependencies

## How has this been tested?
Added tests for the same.

## Checklist


src/sympc/protocol/falcon/falcon.py
```diff
@@ -15,11 +16,15 @@
...
 from sympc.session import Session
 from sympc.session import get_session
+from sympc.store import CryptoPrimitiveProvider
+from sympc.store.exceptions import EmptyPrimitiveStore
 from sympc.tensor import MPCTensor
 from sympc.tensor import ReplicatedSharedTensor
 from sympc.tensor.tensor import SyMPCTensor
 from sympc.utils import parallel_execution
 
+shares_sum = ReplicatedSharedTensor.shares_sum
```
aanurraj, line 26: Q: what does this do?

rasswanth-s: Considering a tensor object ,it might have different operators for summation(binary-xor , normal rings = add(+),prime = add(+) mod prime,it sums the shares according to the ring_size input.

rasswanth-s: In several places for operation on tensors of RSTensor, we call this method.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -105,14 +107,22 @@ def mul_master(
 
         result = None
 
+        ring_size = int(x.share_ptrs[0].get_ring_size().get_copy())
```
gmuraru, line 110: Q: Why not take the `ring_size` from `MPCTensor`, `session` attribute?

rasswanth-s: We operate in three ring_size (2,67,session_rings(2**64,2**32)), an MPC Tensor might be either of three, the ring_size is needed here for beaver triples to generate binary shares, and in truncation and several other places . We are moving taking all the config info at the tensor level instead of session level.This approach reduced several re-encoding issues and makes it compatible with various ring_sizes.

gmuraru: Q: But the ```get_ring_size``` where it is in the ```RST``` Tensor? (I do not see it here)

src/sympc/api.py
```diff
@@ -227,6 +227,14 @@
         "sympc.tensor.replicatedshare_tensor.ReplicatedSharedTensor.numel",
         "syft.lib.python.Int",  # FIXME: Can't we just return an int??
     ),
+    (
+        "sympc.tensor.replicatedshare_tensor.ReplicatedSharedTensor.get_ring_size",
```
gmuraru, line 231: MAYBE: This one should be added to the ReplicatedShareTensor?

src/sympc/protocol/beaver/beaver.py
```diff
@@ -98,25 +102,39 @@ def _get_triples(
...
 
-        a_val = sum([a_shares[0].shares[0]] + a_shares[1].shares)
-        b_val = sum([b_shares[0].shares[0]] + b_shares[1].shares)
-        c_val = sum([c_shares[0].shares[0]] + c_shares[1].shares)
+        shares_sum = ReplicatedSharedTensor.shares_sum
+        a_val = shares_sum([a_shares[0].shares[0]] + a_shares[1].shares, ring_size)
+        b_val = shares_sum([b_shares[0].shares[0]] + b_shares[1].shares, ring_size)
+        c_val = shares_sum([c_shares[0].shares[0]] + c_shares[1].shares, ring_size)
+
+        op = ReplicatedSharedTensor.get_op(ring_size, op_str)
```
gmuraru, line 135: Q: Why do you need the ```get_op```? Can't you directly do?

```
def __add__(self, other):
     if self.ring_size == specific_size:
        # Special add
     else:
        # normal add
```

src/sympc/protocol/falcon/falcon.py
rasswanth-s: https://github.com/OpenMined/SyMPC/pull/269/files#diff-41d43b95b2abaf26d7dffd507ee7a5755e1fcd044e55e0b0485669e8ff9fc622R146

src/sympc/api.py
rasswanth-s: Crct.It here george: 
https://github.com/OpenMined/SyMPC/blob/113895e1e15b253b2e7467a05cebf0f2876e139d/src/sympc/tensor/replicatedshare_tensor.py#L146

src/sympc/protocol/beaver/beaver.py
rasswanth-s: Good ques. For RST addition,sub,mul ,we could use the same ops(`+,-,*`),based on the ring_size ,it would  add.There are scenarios where need to do ops directly on torch.tensor(beaver triples, triple verification,reconstruction...). I initially thought to override torch method,but as you suggested in slack, I thought not to do it.I kept the ops retrieval for  tensors in a central location(i.e RSTensor) ,so we could dynamically add ops on requirement.
https://github.com/OpenMined/SyMPC/blob/113895e1e15b253b2e7467a05cebf0f2876e139d/src/sympc/tensor/replicatedshare_tensor.py#L240

src/sympc/protocol/aby3/aby3.py
```diff
@@ -109,15 +123,36 @@ def truncate(x: MPCTensor, session: Session) -> MPCTensor:
...
         ptr_list = []
         ptr_name = x.share_ptrs[0].__name__
+
+        # TODO:Shoud be concised,lot of branching done,to improve communication efficiency.
+
         if ptr_name == "ReplicatedSharedTensorPointer":
-            ptr_list.append(x.share_ptrs[0].get_shares()[0].get_copy())
-            ptr_list.extend(x.share_ptrs[1].get_copy().shares)
+
+            if ring_size in {2, PRIME_NUMBER}:
```
gmuraru, line 131: Q: Can't the if from here and the `if` from `elif` be in a single `if...else` outside those?

src/sympc/protocol/aby3/aby3.py
```diff
@@ -109,15 +123,36 @@ def truncate(x: MPCTensor, session: Session) -> MPCTensor:
...
+                share_ptrs = x.share_ptrs
```
gmuraru, line 132: Q: For 2 we do not need truncation right? but why don't we need for `PRIME_NUMBER`?

src/sympc/session/session.py
```diff
@@ -232,31 +241,41 @@ def przs_generate_random_share(
...
                 session_uuid=self.uuid,
                 config=Config(encoder_precision=0),
+                ring_size=ring_size,
             )
         return share
 
     def prrs_generate_random_share(
         self,
         shape: Union[tuple, torch.Size],
+        ring_size: str = None,
```
gmuraru, line 272: MAYBE: `Optional[str]`

src/sympc/session/session.py
```diff
@@ -186,44 +186,53 @@ def get_protocol(self) -> Protocol:
...
         tensor_share2 = generate_random_element(
-            tensor_type=self.tensor_type,
+            tensor_type=tensor_type,
             generator=gen2,
             shape=shape,
         )
 
-        return [tensor_share1, tensor_share2]
+        # PRIME_NUMBER ring share is generated in torch.uint8 datatype
+        if ring_size == PRIME_NUMBER:
```
gmuraru, line 221: Q: Shouldn't we take into consideration also the case when `ring_size` is 2

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Good Ques, binary values we do not need encoding,as the input is only an integral value{0,1}. For Prime order rings,we use it to create mainly secret share of bits {0,1} ,so we do not need encoding.Even when do bit injection of binary bit to general ring size(`2**64,2***32`) we do not need encoding as it is sharing of bits, so we set the (`encoder_precision=0`) .This idea is the crux of the PR, where we take config info from the underlying tensors to automatically prevent truncation.

rasswanth-s: NOTE: We take the info to truncate or not ,only from the first operand.I also use this to prevent re-encoding in select shares.
https://github.com/OpenMined/SyMPC/blob/45274eeb98f297dac04067226b2b801a1072a58a/src/sympc/protocol/falcon/falcon.py#L439

src/sympc/session/session.py
rasswanth-s: For binary ring size the tensor type maps ---> torch.bool,so only binary values are generated, for prime rings tensor type maps--->torch.uint8 `[0-255]` ,use modulus to scale to `[0,67]`

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -122,6 +143,125 @@ def get_shares(self) -> List[torch.Tensor]:
...
+            x(torch.Tensor): input tensor
+            y(torch.tensor): input tensor
+
+        Returns:
+            value(torch.Tensor): Result of the operation.
+
+        Raises:
+            ValueError : If either of the tensors datatype is not torch.uint8
+        """
+        if x.dtype != torch.uint8 or y.dtype != torch.uint8:
```
kamathhrishi, line 178: Why only uint8, why not uint16 or uint32?

kamathhrishi: Same for other ops.

rasswanth-s: For PRIME_NUMBER order ring we tend to use small prime to decrease the share size to be communication efficient,Falcon mentions `37` , but for  practical implementation ,Falcon uses `67`. The prime_order ring is mapped to `torch.uint8` datatype.
https://github.com/OpenMined/SyMPC/blob/18a52383d0c72834c9825dc5d504d07c4e1ebd86/src/sympc/utils/mpc_utils.py#L16

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -519,6 +669,24 @@ def ne(self, y):
...
+
+        Args:
+            shares (List[torch.Tensor]) : List of tensors.
+            ring_size (int): Ring size of share associated with the tensors.
+
+        Returns:
+            value (torch.Tensor): sum of the tensors.
+        """
+        if ring_size == 2:
+            return reduce(lambda x, y: x ^ y, shares)
```
kamathhrishi, line 684: Hmmm, this will be without carry right so it will be inaccurate. Should we defer it to another PR? or raise a warning?

rasswanth-s: The carry part is needed for local decomposition and composition for consistent shares, here we are computing a single binary tensor, we would not need it.

kamathhrishi (approved):
Looks good to me.

src/sympc/protocol/aby3/aby3.py
gmuraru: I will ping you tomorrow about this. Wouldn't ring "2" offer he possibility to have shares of bits?

src/sympc/session/session.py
```diff
@@ -186,44 +186,53 @@ def get_protocol(self) -> Protocol:
...
+            return [tensor_share1 % ring_size, tensor_share2 % ring_size]
```
gmuraru, line 222: I am not sure If doing this generates uniformly random numbers between 0 and 67.
For example, if we generate random numbers between 0 and 4 (inclusive) and this is the `tensor_share` and now we want to generate random numbers between 0 and 1 we would do a `module` with 2.

In this case, we would end up with a number from `(random number from [0, 1, 2, 3, 4]) % 2`, but if you take a closer look you see that we have 3 even numbers (0, 2, 4) and 2 odd numbers (1 and 3) -- so we get numbers in [0, 1] but we tend to get 0 more often than 1.

I am not sure if it is the same case here.

**LE**: Thinking more into this, it might not affect us that much, but we still need to take into consideration if this is really uniformly distributed (Ping me on slack if you have questions)

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Sure George :+1:

src/sympc/session/session.py
rasswanth-s: Good catch george , I will work on it :+1:  ,

rasswanth-s: Generating non uniform random numbers,is a big security flaw, I thought to inspect at it after initial devlopment, a very good catch george, I have fixed the same setting max value in torchcsprng.

gmuraru (approved):
LGTM!

gmuraru:
@rasswanth-s are there any more changes required? We can merge this, right?

PR #274, opened 2021-07-11
Falcon : Private compare

### LINKED ISSUE #148: Falcon Private Compare
## Feature Description
This function evaluates the bit x > r where r is public and the parties hold shares of bits of x in ring Zp
Refer to Algorithm 1 in the paper. 

## Additional Context
[Falcon Paper](https://arxiv.org/abs/2004.02229). This PR has a dependency on #145. 
Task list
- [x] Semi Honest
- [x] Malicious

rasswanth-s:
## Description
Implements the Falcon Private Compare Functionality. fixes #148 

## How has this been tested?
Added tests for the same.

## Checklist


kamathhrishi (commented):
Have a high-level interface for MPCTensor to allow users to perform MPC_Tensor_X>MPC_Tensor_Y or any such comparison operation. Similar to current FSS implementation. Basically, add this to le and gt operation function.

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -289,3 +291,28 @@ def test_select_shares_exception_shape(get_clients) -> None:
...
+def test_private_compare(get_clients, security) -> None:
+    parties = get_clients(3)
+    falcon = Falcon(security_type=security)
+    session = Session(parties=parties, protocol=falcon)
+    SessionManager.setup_mpc(session)
+    base = session.config.encoder_base
+    precision = session.config.encoder_precision
+    fp_encoder = FixedPointEncoder(base=base, precision=precision)
+
+    secret = torch.tensor([[358, 79], [67, 2415]])
```
kamathhrishi, line 306: Why only integers? Doesn't it support float?

rasswanth-s: It works for float , forgot to add them, will add it. :+1:

rasswanth-s:
> Have a high-level interface for MPCTensor to allow users to perform MPC_Tensor_X>MPC_Tensor_Y or any such comparison operation. Similar to current FSS implementation. Basically, add this to le and gt operation function.

Falcon private compare is specific for comparing with public value, We could make it works for mpc tensors also, we have to change the building of blocks of Private Compare.I thought not to expose at mpc tensor level for now. Maybe when we modify the algorithm, we could include at MPCTensor level ?

tests/sympc/protocol/falcon/falcon_test.py
rasswanth-s: Done

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,87 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+            r (torch.Tensor) : Public integer r.
+
+        Returns:
+            result (MPCTensor): Returns shares of bits of the operation.
+
+        (if (x>=r) returns 1 else returns 0)
+        """
+        shape = x[0].shape
+        session = x[0].session
+        ptr_list: List = []
```
gmuraru, line 510: MAYBE:
```
ptr_list = [session_ptr.prrs_generate_random_share(shape=shape, ring_size="2") for session_ptr in session.session_ptrs]
```

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,87 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+
+        beta_2 = MPCTensor(
+            shares=ptr_list, session=session, shape=shape
+        )  # shares of random bit
+        beta_p = ABY3.bit_injection(
+            beta_2, session, PRIME_NUMBER
+        )  # shares of random bit in Zp.
+        m = Falcon._random_prime_group(session, shape)
+
+        u = [0 for i in range(len(x))]
```
gmuraru, line 524: MAYBE: Save `len(n)` as `nr_shares = len(n)`
Then:
```
u = [0] * nr_shares
w = [0] * nr_shares
c = [0] * nr_shares
```

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,87 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+        w = [0 for i in range(len(x))]
+        c = [0 for i in range(len(x))]
+        for i in range(len(x) - 1, -1, -1):
+            r_i = r >> i & 1  # bit at ith position
+            u[i] = (1 - 2 * beta_p) * (x[i] - r_i)
+            w[i] = x[i] ^ r_i
+            c[i] = u[i] + 1 + sum(w[i + 1 :])
```
gmuraru, line 531: Q: Isn't a bug here? when `i == len(x) - 1` we do `w[len(x) :])` but `w` can be indexed from `0 to len(x) - 1`

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,87 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+
+        Args:
+            session (Session): session to generate random shares for.
+            shape (Union[torch.Size,tuple]): shape of the random share to generate.
+
+        Returns:
+            share (MPCTensor): Retuns shares of random number in group Zp*.
+
+        Zp* = {1,2..,p-1},where p is a prime number.
+        We use Euler's Theorum for verifying that random share is not zero.
```
gmuraru, line 472: TYPO: Euler's Theorem

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,87 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+        a^phi(n) = 1(mod n), if a is co prime to n.
+        In our case n=p(prime number), phi(p) = p-1
+        phi(n) = Euler totient function.
+        """
+        while True:
+            ptr_list: List[ReplicatedSharedTensor] = []
+            for session_ptr in session.session_ptrs:
+                ptr = session_ptr.prrs_generate_random_share(
+                    shape=(), ring_size=str(PRIME_NUMBER)
+                ).resolve_pointer_type()
```
gmuraru, line 484: Q: Can't we remove the resolve pointer here?

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Good question, george, I will effectively wanted it to return 0 ,when `i==len(x)-1` , when we slice `w[i+1:]` ,it returns an empty list, when we use sum on top of empty,it returns zero,which fit the scenario.

src/sympc/protocol/falcon/falcon.py
rasswanth-s: It returns an `Any Pointer` after `repeat` method,which could not be used in subsequent operations, so I used `resolve_pointer_type`

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Done

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Done

src/sympc/protocol/falcon/falcon.py
gmuraru: But it should be good I think. `Any` pointer is only for us to know what methods are on that pointer, but you should be able to still use that pointer I think.
Q: Could you remove the `resolve_pointer_type` and check if it works without it?

src/sympc/protocol/falcon/falcon.py
gmuraru: Ahh...Nice - interesting behavior for python - I will recheck this logic today.

src/sympc/protocol/falcon/falcon.py
rasswanth-s: I tried removing it george ,I got an error ,as we subscript share ptrs in several places of the underlying tensor.

kamathhrishi: Also that I think PRRS has a Union of RSTensor and ShareTensor in API.py so typically such methods require resolve_pointer_type().

src/sympc/protocol/falcon/falcon.py
gmuraru: MAYBE: Can't we keep `w` as a single number?
```
w = 0
for i in range(len(x) - 1, -1, -1):
            r_i = r >> i & 1  # bit at ith position
            u[i] = (1 - 2 * beta_p) * (x[i] - r_i)
            w += x[i] ^ r_i
            c[i] = u[i] + 1 + w
```

In this case, you won't have to always do the `sum`.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,89 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+        )  # shares of random bit in Zp.
+        m = Falcon._random_prime_group(session, shape)
+
+        nr_shares = len(x)
+        u = [0] * nr_shares
+        w = [0] * nr_shares
+        c = [0] * nr_shares
+
+        for i in range(len(x) - 1, -1, -1):
+            r_i = r >> i & 1  # bit at ith position
```
gmuraru, line 530: MAYBE: Could you also put `(...)` around the first operation. Sometimes is not that intuitive to see what is the first one that is executed.

src/sympc/protocol/falcon/falcon.py
rasswanth-s: You mean something like also sanity check george, like when
<code>if i==len(x)-1 <br>
w=0
</code>

rasswanth-s: Crct George,  running sum is inefficient in the context of MPCTensor, I would keep and current variable at end and change it

src/sympc/protocol/falcon/falcon.py
rasswanth-s: George, in Any pointer , the tensor specific methods are not copied I think , which resulted in error at Falcon multiplication, which used tensor specific methods of RSTensor.

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Modified.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,100 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+            ValueError: If input shares is not a list.
+            ValueError: If input public value is not a tensor.
+
+        (if (x>=r) returns 1 else returns 0)
+        """
+        if not isinstance(x, list):
+            raise ValueError(f"Input shares for Private Compare: {x} must be a list")
+
+        if not isinstance(r, torch.Tensor):
+            raise ValueError(f"Value r:{r} must be a tensor for private compare")
```
gmuraru, line 516: NITPICK: Could you specify here torch tensor?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,100 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+
+        shape = x[0].shape
+        session = x[0].session
+
+        ptr_list: List[ReplicatedSharedTensor] = [
+            session_ptr.prrs_generate_random_share(shape=shape, ring_size=str(2))
```
gmuraru, line 522: NITPICK: `ring_size="2"`

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,100 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+
+        w = 0
+
+        for i in range(len(x) - 1, -1, -1):
+            r_i = (r >> i) & 1  # bit at ith position
+            u[i] = (1 - 2 * beta_p) * (x[i] - r_i)
+            c[i] = u[i] + 1 + w
+            w += x[i] ^ r_i
+
+        d = m * (math.prod(c))
```
gmuraru, line 546: NITPICK: No need here to pun `(...)` around `math.prod`. We know that a function call is done first

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,100 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+
+    @staticmethod
+    def _random_prime_group(
+        session: Session, shape: Union[torch.Size, tuple]
+    ) -> MPCTensor:
+        """Computes shares of random number in Zp*.Zp* is the multiplicative group mod p.
+
+        Args:
+            session (Session): session to generate random shares for.
+            shape (Union[torch.Size,tuple]): shape of the random share to generate.
```
gmuraru, line 466: NITPICK: one empty space between `tuple` and `,`

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,100 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+
+        Returns:
+            share (MPCTensor): Returns shares of random number in group Zp*.
+
+        Zp* = {1,2..,p-1},where p is a prime number.
+        We use Euler's Theorem for verifying that random share is not zero.
+        It states that:
+        For a general modulus n
+        a^phi(n) = 1(mod n), if a is co prime to n.
```
gmuraru, line 475: Q: Why do we need this?
Can't we simply generate random shares in ring size `p` on each party and when reconstructing they give back a random number?

rasswanth-s: Good question, we do this to prevent the random share from being zero,

rasswanth-s: I will include as comment also.

rasswanth-s: We try effectively to sample from the distribution [1,p-1]

rasswanth-s: We cannot restrict parties to sample in a specific manner(which is a security leak), we allow them to sample from domain [0,p-1] and check if the shares sum to zero.

rasswanth-s: We do it mainly to mask the c values, if the random shares is zero, it reveals the d output implicitly on reconstruction.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,102 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+
+            m = MPCTensor(shares=ptr_list, session=session, shape=shape)
+
+            m_euler = m ** (PRIME_NUMBER - 1)
+
+            if (m_euler.reconstruct(decode=False) == 1).all():
+                return m
+
+    @staticmethod
+    def private_compare(x: List[MPCTensor], r: torch.Tensor) -> MPCTensor:
```
gmuraru, line 498: Q: Should we do any check to make sure that the `r` has at least `ring_size` bits?

rasswanth-s: By ring_size you mean the session tensor type geoge

rasswanth-s: Currently we place the onus on the user to provide the correct tensor type.

rasswanth-s: Including a sanity check for r would be good :+1:  , I would include it.

rasswanth-s: I think we do not need a sanity check, in retrospection, it works for arbitrary tensor types also , as we do an internal type conversion in RSTensor apply_op,

rasswanth-s: If we are comparing 64 bit number with  8 bit r values, the positions at [9,63] would be zero, which yields the correct result

gmuraru: Ahh...it is fine. I was thinking that it is `undefined behaviour` for shifting (to left/right) more than the number of bits for that data type.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,102 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
...
+        In our case n=p(prime number), phi(p) = p-1
+        phi(n) = Euler totient function.
+        We effectively try to sample a random number in range [1,p-1],discard the instances where
+        it equals zero.
+        """
+        while True:
+            ptr_list: List[ReplicatedSharedTensor] = []
+            for session_ptr in session.session_ptrs:
+                ptr = session_ptr.prrs_generate_random_share(
+                    shape=(), ring_size=str(PRIME_NUMBER)
```
gmuraru, line 485: Q: Why not directly specify the shape here and you call `repeat`?

gmuraru (approved):
LGTM! :rocket: only that question Good job!

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Discussed on slack!

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -289,3 +291,42 @@ def test_select_shares_exception_shape(get_clients) -> None:
...
+
+
+def test_private_compare_exception(get_clients) -> None:
+    parties = get_clients(3)
+    falcon = Falcon()
+    session = Session(parties=parties, protocol=falcon)
+    SessionManager.setup_mpc(session)
+    x = MPCTensor(secret=1, session=session)
+    r = torch.tensor([1])
+    with pytest.raises(ValueError):
```
kamathhrishi, line 328: Could you add a comment as why to an exception will be raised here?

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -289,3 +291,42 @@ def test_select_shares_exception_shape(get_clients) -> None:
...
+        Falcon.private_compare(x, r)
+
+    with pytest.raises(ValueError):
```
kamathhrishi, line 331: Same

src/sympc/protocol/falcon/falcon.py
```diff
@@ -451,3 +454,102 @@ def select_shares(x: MPCTensor, y: MPCTensor, b: MPCTensor) -> MPCTensor:
         z = x + (d * (y - x))
 
         return z
+
+    @staticmethod
+    def _random_prime_group(
```
kamathhrishi, line 459: This can be extended beyond private compare right? 
Maybe we could have a test for this?

rasswanth-s: Crct ,it could be extended beyond PC compare.
I think we should not test private methods Hrishi
https://softwareengineering.stackexchange.com/questions/100959/how-do-you-unit-test-private-methods/100966
We test it by testing Private Compare.

tests/sympc/protocol/falcon/falcon_test.py
rasswanth-s: Done

tests/sympc/protocol/falcon/falcon_test.py
rasswanth-s: Done

kamathhrishi (approved):

PR #280, opened 2021-07-18
Allow additional attributes from torch modules into SyMPC modules

kamathhrishi:
## Description
Currently, only weights and bias have been copied from torch modules to sympc modules (conv2d and Linear). But, additional attributes as in case of conv2d such as padding and stride haven't been copied. This results in differing functionality from intended behavior.

## How has this been tested?
- Modified test to include stride 

**Reproducing Bug** 
This could be reproduced in the current version of the codebase by modifying https://github.com/OpenMined/SyMPC/blob/090c11dbefbebaab160c6480ade2d3df120715a1/tests/sympc/module/module_test.py#L37 to include a stride, padding or any value that is different from the default (say 2). Further, modifying the linear layer dimensions accordingly. You will notice that test fails because behaviour deviates from intended behaviour (as done by torch on a given input). 

## Checklist


tests/sympc/module/module_test.py
```diff
@@ -34,15 +34,15 @@ class ConvNet(sy.Module):
     def __init__(self, torch_ref, kernel_size=5):
```
gmuraru, line 34: Q: Could you validate that stride was added to the module?

kamathhrishi: Done.

src/sympc/module/nn/conv.py
```diff
@@ -78,46 +80,65 @@ def forward(self, x: MPCTensor) -> MPCTensor:
...
         # we have groups == 1
+
         (
             self.out_channels,
             self.in_channels,
             kernel_size_w,
             kernel_size_h,
         ) = shape
 
-        if kernel_size_w != kernel_size_h:
```
kamathhrishi, line 114: Allow rectangular kernels.

src/sympc/module/nn/functional.py
```diff
@@ -262,6 +262,9 @@ def max_pool2d(
     Raises:
         ValueError: if the kernel size is bigger than the input
     """
+    if x.session.nr_parties != 2:
```
kamathhrishi, line 265: Added check to ensure only 2 parties are present in session.

gmuraru (approved):
LGTM!

PR #284, opened 2021-07-24
Add GradDiv to Autograd 

kamathhrishi:
## Description
This PR adds GradDiv functionality. 
Also includes small changes such as 
1) Adding parameters function to Conv2d to allow parameters to be passed to the optimizer
2) Added MPCTensor reciprocal function with test 

## How has this been tested?
Added unit tests for the same. 

## Checklist


src/sympc/api.py
```diff
@@ -247,6 +247,10 @@
         "sympc.tensor.replicatedshare_tensor.ReplicatedSharedTensor.view",
         "sympc.tensor.replicatedshare_tensor.ReplicatedSharedTensor",
     ),
+    (
+        "sympc.tensor.replicatedshare_tensor.ReplicatedSharedTensor.reshape",
```
gmuraru, line 251: Q: I don't think you need this, right?

kamathhrishi: Not for this pr. I was just adding in general, because it will be used in future. I found the use when reshaping outputs for neural networks and thought I should add it.

gmuraru: Could you remove it from here and we can add it when we need it?

src/sympc/grads/grad_functions.py
```diff
@@ -351,6 +351,44 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> MPCTensor:
...
+        Args:
+            ctx (Dict[str, Any]): Context used to retrieve the information for the backward pass
+            grad (MPCTensor): The gradient that came from the child nodes
+
+        Returns:
+            (x_grad, y_grad) (Tuple[MPCTensor]): The gradients passed to the X and Y nodes.
+        """
+        x, y = ctx["x"], ctx["y"]
+
+        grad_x = grad * (x / (y * x))
```
gmuraru, line 386: Q: Why does it look like this?
```
dx x/y = 1/y
```
So for this, you will have `grad / y `.

src/sympc/grads/grad_functions.py
```diff
@@ -351,6 +351,44 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> MPCTensor:
...
+        grad_y = ((-1) * (x)) / ((y ** 2) * grad)
```
gmuraru, line 387: MAYBE: Why not save in the forward the `x / y` and then here multiply by `(-1 * x * grad_cache * grad)`

tests/sympc/grad/grad_functions_test.py
```diff
@@ -814,3 +815,53 @@ def test_grad_maxpool_2d_backward(get_clients, kernel_size, stride, padding) ->
...
+    y_secret = torch.tensor([9.0, 8.0])
+    y = MPCTensor(secret=y_secret, session=session, requires_grad=True)
+
+    grad_mpc = MPCTensor(secret=grad, session=session, requires_grad=True)
+
+    ctx = {"x": x, "y": y}
+
+    grad_x, grad_y = GradDiv.backward(ctx, grad_mpc)
+
+    expected_grad_x = grad * (x_secret / (y_secret * x_secret))
```
gmuraru, line 860: Can you test here with torch? (like do division and on the resulting tensor do the backward pass and it should have approximately the same result)

tests/sympc/grad/grad_functions_test.py
```diff
@@ -814,3 +815,53 @@ def test_grad_maxpool_2d_backward(get_clients, kernel_size, stride, padding) ->
...
+    expected_grad_y = ((-1) * (x_secret)) / ((y_secret ** 2) * grad)
+
+    res_x = grad_x.reconstruct()
+    res_y = grad_y.reconstruct()
+
+    assert np.allclose(res_x, expected_grad_x, rtol=1e-1)
```
gmuraru, line 866: Q: Can we have here `1e-2`

gmuraru (changes requested):
See comments

src/sympc/grads/grad_functions.py
aanurraj: I dont understand this diff :/

src/sympc/module/nn/conv.py
```diff
@@ -140,11 +141,25 @@ def share_state_dict(
...
         self.weight = MPCTensor(secret=weight, session=self.session, shape=shape)
+        self._parameters = OrderedDict({"weight": self.weight})
 
         if bias is not None:
             self.bias = MPCTensor(
                 secret=bias, session=self.session, shape=(self.out_channels,)
             )
+            self._parameters["bias"] = self.bias
+
+    def parameters(self, recurse: bool = False) -> MPCTensor:
```
aanurraj, line 152: Q: Does this solves the None issue in step while training?

kamathhrishi: No, this doesn't. The whole GradDiv PR solves that issue. Basically, MSELoss has a division operation. GradDiv hasn't defined some gradients after the division was None.

tests/sympc/grad/grad_functions_test.py
kamathhrishi: Yeah sure. But, I won't just do (x/y) and take gradient. More like (x/y).sum() and take gradient because (x/y) is not a scalar. Anyway, sum has gradient 1.

tests/sympc/grad/grad_functions_test.py
kamathhrishi: The tests were failing because there were small differences and the numbers are also small. The output in plaintext was: _tensor([-0.0268, -0.0234])_ and the result after reconstruction:  _tensor([-0.0247, -0.0234])_. Instead of keeping it this way I will change the values being passed (x,y, grad) so that gradients are small enough and we can use 1e-2.

src/sympc/grads/grad_functions.py
kamathhrishi: Good question. I have no idea :P but I am figuring it out. 

I know grad(x/y) wrt to x is (1/x). 
When I first made it (1/y) and compared results with PyTorch for the same, they were very different. But, with some random experimentation I found that (x/(x*y) matches Pytorch's grad for division.

src/sympc/api.py
kamathhrishi: Alright.

tests/sympc/grad/grad_functions_test.py
kamathhrishi: Nevermind.

src/sympc/grads/grad_functions.py
kamathhrishi: Nevermind this. I will correct it. I made a mistake.

src/sympc/grads/grad_functions.py
```diff
@@ -351,6 +351,48 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> MPCTensor:
...
+            ctx (Dict[str, Any]): Context used to retrieve the information for the backward pass
+            grad (MPCTensor): The gradient that came from the child nodes
+
+        Returns:
+            (x_grad, y_grad) (Tuple[MPCTensor]): The gradients passed to the X and Y nodes.
+        """
+        x, y = ctx["x"], ctx["y"]
+
+        grad_x = grad * (1 / y)
+        grad_y = (-1) * grad * (ctx["result"] / y)
```
rasswanth-s, line 390: MAYBE : grad_y = (-1)*grad_x *ctx["result"]
We could save a division operation,I think division is compute intensive due to approximations.

src/sympc/grads/grad_functions.py
rasswanth-s: Q:We should remove the previous method?

src/sympc/grads/grad_functions.py
```diff
@@ -135,6 +135,12 @@ def forward(ctx: Dict[str, Any], x: MPCTensor, y: Any) -> MPCTensor:
         Returns:
             x + y (MPCTensor): The result of the addition
         """
+        if not hasattr(x, "shape"):
```
gmuraru, line 138: Insert meme with "we do not do that here"

gmuraru: Can't you leave it only for the tensor case at the moment? this should be treated better

src/sympc/module/nn/conv.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, session: Session) -> None:
         self.padding = 0
         self.dilation = 1
         self.groups = 1
+        self._parameters = None
```
gmuraru, line 60: Q: Could you add those to slots and change the slots to a tuple from the list?

src/sympc/grads/grad_functions.py
kamathhrishi: Nope, it can't be only for tensor case. The reciprocal function uses addition with an integer. So this throws an error saying _"Integer has no attribute shape"_

src/sympc/module/nn/conv.py
kamathhrishi: Yeah sure.

src/sympc/grads/grad_functions.py
kamathhrishi: I can just cast the integers in reciprocal as torch tensors and add it instead.

kamathhrishi: Done

src/sympc/module/nn/conv.py
kamathhrishi: Done

src/sympc/approximations/reciprocal.py
```diff
@@ -31,7 +34,7 @@ def reciprocal(self, method: str = "NR", nr_iters: int = 10):
 
     if method == "nr":
         new_self = modulus(self)
-        result = 3 * exp(0.5 - new_self) + 0.003
+        result = 3 * exp(0.5 - new_self) + torch.tensor(0.003)
```
gmuraru, line 37: Q: Can't you remove the tensor from here and do it in the "GradDiv":
```

if isinstance(x, (list, int, float)):
   x = torch.tensor(x)
```

tests/sympc/grad/grad_functions_test.py
```diff
@@ -814,3 +815,55 @@ def test_grad_maxpool_2d_backward(get_clients, kernel_size, stride, padding) ->
...
+    session.autograd_active = True
+    SessionManager.setup_mpc(session)
+
+    x_secret = torch.tensor([1.0, 2.0, 3.0, 4.0], requires_grad=True)
+    x = MPCTensor(secret=x_secret, session=session, requires_grad=True)
+
+    y_secret = torch.tensor([2.0, 3.0, 4.0, 5.0], requires_grad=True)
+    y = MPCTensor(secret=y_secret, session=session, requires_grad=True)
+
+    z = (x_secret / y_secret).sum()
```
gmuraru, line 852: OUGHTTO: Can you remove the `sum` from here?
For the moment this test should check only the `div` functionality (you can pass in the gradient when doing `backward`)

tests/sympc/tensor/mpc_tensor_test.py
```diff
@@ -787,3 +787,17 @@ def test_session_ring_xor(get_clients, security, bit) -> None:
     result = operator.xor(x, b)
     expected_res = secret_x ^ secret_b
     assert (result.reconstruct(decode=False) == expected_res).all()
+
+
+def test_mpctensor_reciprocal(get_clients):
```
gmuraru, line 792: MAYBE: Could you leave only `test_reciprocal` no need to specify mpc tensor since it is in the `mpc tensor file`

src/sympc/approximations/reciprocal.py
kamathhrishi: In gradsum you mean?

tests/sympc/grad/grad_functions_test.py
kamathhrishi: Done

tests/sympc/tensor/mpc_tensor_test.py
kamathhrishi: Done

src/sympc/approximations/reciprocal.py
kamathhrishi: Done

src/sympc/grads/grad_functions.py
```diff
@@ -135,6 +135,12 @@ def forward(ctx: Dict[str, Any], x: MPCTensor, y: Any) -> MPCTensor:
         Returns:
             x + y (MPCTensor): The result of the addition
         """
+        if isinstance(x, (list, int, float)):
```
gmuraru, line 138: MAYBE: I think you should remove this because it reached this point because x was an MPCTensor, right?

src/sympc/grads/grad_functions.py
```diff
@@ -243,6 +252,12 @@ def forward(ctx: Dict[str, Any], x: MPCTensor, y: Any) -> MPCTensor:
         Returns:
             x - y (MPCTensor): The result of the substraction
         """
+        if isinstance(x, (list, int, float)):
```
gmuraru, line 255: Also here

kamathhrishi: Right will remove. Dint realize.

src/sympc/grads/grad_functions.py
```diff
@@ -281,9 +290,10 @@ def forward(ctx: Dict[str, Any], x: MPCTensor, y: Any) -> MPCTensor:
         Returns:
             x * y (MPCTensor): The result of the multiplication
         """
-        if not hasattr(x, "shape"):
+        if isinstance(x, (list, int, float)):
```
gmuraru, line 293: You should check only `y`

src/sympc/grads/grad_functions.py
```diff
@@ -351,6 +361,50 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> MPCTensor:
...
+        Args:
+            ctx (Dict[str, Any]): Context used to retrieve the information for the backward pass
+            grad (MPCTensor): The gradient that came from the child nodes
+
+        Returns:
+            (x_grad, y_grad) (Tuple[MPCTensor]): The gradients passed to the X and Y nodes.
+        """
+        y = ctx["y"]
+
+        grad_x = grad * (1 / y)
```
gmuraru, line 402: Q: Can't you do `grad / y`?

tests/sympc/grad/grad_functions_test.py
```diff
@@ -814,3 +815,55 @@ def test_grad_maxpool_2d_backward(get_clients, kernel_size, stride, padding) ->
...
+    x = MPCTensor(secret=x_secret, session=session, requires_grad=True)
+
+    y_secret = torch.tensor([2.0, 3.0, 4.0, 5.0], requires_grad=True)
+    y = MPCTensor(secret=y_secret, session=session, requires_grad=True)
+
+    z = x_secret / y_secret
+    z.backward(torch.tensor([1, 1, 1, 1]))
+
+    grad = torch.tensor([1, 1, 1, 1])
+    grad_mpc = MPCTensor(secret=grad, session=session, requires_grad=True)
```
gmuraru, line 856: :+1: :+1: :+1:

src/sympc/grads/grad_functions.py
```diff
@@ -173,6 +176,9 @@ def forward(ctx: Dict[str, Any], x: MPCTensor) -> MPCTensor:
         Returns:
             sum(x) (MPCTensor): The summation of all the elements from the input
         """
+        if isinstance(x, (list, int, float)):
```
rasswanth-s, line 179: As george mentioned in previous comments, should we omit checking for x here?

src/sympc/grads/grad_functions.py
```diff
@@ -351,6 +361,50 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> MPCTensor:
...
+
+        Args:
+            ctx (Dict[str, Any]): Context used to save information needed in the backward pass
+            x (MPCTensor): 1st operand for the division operation
+            y (Any): 2nd operand for the divison operation
+
+        Returns:
+            x / y (MPCTensor): The result of the division
+        """
+        if isinstance(x, (list, int, float)):
```
rasswanth-s, line 379: Q: same question ,Should we check for x here?

src/sympc/tensor/mpc_tensor.py
```diff
@@ -576,6 +576,18 @@ def rmatmul(self, y: torch.Tensor) -> "MPCTensor":
 
         return result
 
+    def rtruediv(self, y: Union["MPCTensor", torch.Tensor, float, int]) -> "MPCTensor":
```
rasswanth-s, line 579: Since it is an `rtruediv`, I think y does not include `MPCTensor`

src/sympc/tensor/mpc_tensor.py
```diff
@@ -576,6 +576,18 @@ def rmatmul(self, y: torch.Tensor) -> "MPCTensor":
...
+        """Apply recriprocal of MPCTensor.
+
+        Args:
+            y (Union["MPCTensor", torch.Tensor, float, int]): Numerator.
```
rasswanth-s, line 583: Here also

tests/sympc/tensor/mpc_tensor_test.py
```diff
@@ -787,3 +787,17 @@ def test_session_ring_xor(get_clients, security, bit) -> None:
...
     expected_res = secret_x ^ secret_b
     assert (result.reconstruct(decode=False) == expected_res).all()
+
+
+def test_reciprocal(get_clients):
+    clients = get_clients(2)
+    session = Session(parties=clients)
+    SessionManager.setup_mpc(session)
+
+    x_secret = torch.Tensor([1, 2, 3, 4])
```
rasswanth-s, line 797: MAYBE: Includes also float values?

rasswanth-s (approved):
LGTM! Some Comments.

tests/sympc/grad/grad_functions_test.py
kamathhrishi: This is fine right?

gmuraru (approved):
LGTM!

PR #287, opened 2021-07-29
[WIP] Falcon : Wrap Function and added support for Numpy in RSTensor.

### LINKED ISSUE #150: Falcon Wrap Function
## Feature Description
Implement the Wrap function mentioned in Falcon Protocol. It is required for RELU and DRELU function implementations. 

## Additional Context
[Falcon Paper](https://arxiv.org/abs/2004.02229). The Issue has a dependency on #145.

rasswanth-s:
## Description
Implementation of Falcon wrap functionality, which is a secure protocol for computing carry bit on addition of three values.resolves #150 

## How has this been tested?
To be Added.

## Checklist


src/sympc/protocol/falcon/falcon.py
```diff
@@ -404,3 +431,289 @@ def multiplication_protocol(
...
+        Returns:
+            result (np.ndarray): Modulo reduction of the exact wrap function.
+
+        Raises:
+            ValueError: If the input values is not a numpy array.
+            ValueError: If signed numpy array are provided as input.
+
+        wrap3 calucaties the carry bit on addition of three values a+b+c(mod 2).
+        """
+        if not isinstance(a, np.ndarray) or not isinstance(b, np.ndarray):
```
aanurraj, line 617: Q: shouldn't we check for c too?

aanurraj (changes requested):

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Yeah we should for c,I would change it :+1:

rasswanth-s: Done

src/sympc/__init__.py
```diff
@@ -25,6 +25,9 @@
 try:
     # third party
     import syft
+    from syft.experimental_flags import flags
+
+    flags.APACHE_ARROW_TENSOR_SERDE = False
```
gmuraru, line 30: Q: This is because syft uses `ARROW` for numpy?

src/sympc/protocol/aby3/aby3.py
```diff
@@ -92,10 +93,25 @@ def truncation_algorithm1(
         base = config.encoder_base
         precision = config.encoder_precision
         scale = base ** precision
+
         x1, x2, x3 = ptr_list
+        # TODO: UNSIGNED Truncation does not work for numpy, should be modified.
+        dtype = ptr_list[0].dtype
+        if isinstance(ptr_list[0], np.ndarray):
```
gmuraru, line 100: Q: Shouldn't it work - if you have a uint64 numpy array type and you want to add another np.uint array to it, it should wrap around, right?
I think the problem comes when you add an integer, and at that point the numpy type gets converted to an `object` type.

src/sympc/__init__.py
rasswanth-s: Crct george,Syft used `ARROW` for numpy. The `serde` is buggy in the case of `ARROW` ,
The de-serialization of boolean type are erroneous.
I opened an issue regarding the same in PySyft george now.
https://github.com/OpenMined/PySyft/issues/5930

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Yes, George, it should wrap around, I am pretty sure the truncation algorithm should work for unsigned types, it needs additional look into  the implementation semantics.I will try figuring out the problem.

Couldn't get you  , the `object` type meaning?

kamathhrishi, line 100: Add NumPy support to type annotation?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -553,3 +585,156 @@ def private_compare(x: List[MPCTensor], r: torch.Tensor) -> MPCTensor:
...
+            ValueError : If the input tensor shape is None.
+        """
+        shape = a.shape
+        if shape is None:
+            raise ValueError("Input MPCTensor for Wrap must have valid shape")
+        tensor_type = session.tensor_type
+        dtype = UNSIGNED_MAP[tensor_type]
+        ring_size = session.ring_size
+
+        ptr_list: List = []
```
kamathhrishi, line 667: Maybe list comprehension ?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -553,3 +585,156 @@ def private_compare(x: List[MPCTensor], r: torch.Tensor) -> MPCTensor:
...
+        ):
+            raise ValueError("Input value must be a numpy array for wrap3.")
+
+        if a.dtype.kind == "i" or b.dtype.kind == "i" or c.dtype.kind == "i":
+            raise ValueError("Wrap3 works only for signed numbers.")
+
+        return Falcon.wrap2(a, b) ^ Falcon.wrap2(a + b, c)
+
+    @staticmethod
+    def wrap_preprocess(a: MPCTensor, session: Session) -> List:
```
kamathhrishi, line 647: Could you change the annotation to List of MPCTensors? Same for docstrings.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -553,3 +585,156 @@ def private_compare(x: List[MPCTensor], r: torch.Tensor) -> MPCTensor:
...
+
+        a, x, x_p, alpha = Falcon.wrap_preprocess(a, session)
+
+        r = x + a
+
+        # TODO : change to get shares by reconstruct,malicious returns all_shares
+        r1 = r.share_ptrs[0].get_copy().shares[0]
+        r2, r3 = r.share_ptrs[1].get_copy().shares
+
+        share_ptrs = []
```
kamathhrishi, line 724: Maybe: List comphrehension?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -553,3 +585,156 @@ def private_compare(x: List[MPCTensor], r: torch.Tensor) -> MPCTensor:
...
+        Returns:
+            wrap_sh (MPCTensor): bit shares of wrap of the input tensor.
+        """
+        session = a.session
+
+        a, x, x_p, alpha = Falcon.wrap_preprocess(a, session)
+
+        r = x + a
+
+        # TODO : change to get shares by reconstruct,malicious returns all_shares
```
kamathhrishi, line 720: Lets toggle based on mode for protocol in session.

src/sympc/protocol/falcon/falcon.py
```diff
@@ -553,3 +585,156 @@ def private_compare(x: List[MPCTensor], r: torch.Tensor) -> MPCTensor:
...
+        r1 = r.share_ptrs[0].get_copy().shares[0]
```
kamathhrishi, line 721: Why not Just add MPCTensors?

tests/sympc/protocol/falcon/falcon_test.py
```diff
@@ -333,3 +333,22 @@ def test_private_compare_exception(get_clients) -> None:
     # Exception for not passing a public value(torch.Tensor).
     with pytest.raises(ValueError):
         Falcon.private_compare([x], x)
+
+
+def test_wrap(get_clients) -> None:
```
kamathhrishi, line 338: Let's add a test for malicious security type?

src/sympc/protocol/aby3/aby3.py
gmuraru: Hmm..a first question  - from where does the unsigned comes? (the unsigned numpy)

rasswanth-s: Crct , the error occurs for unsigned numpy types.

tests/sympc/protocol/falcon/falcon_test.py
rasswanth-s: Added

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Modified

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Done

src/sympc/protocol/falcon/falcon.py
rasswanth-s: Done

src/sympc/protocol/aby3/aby3.py
rasswanth-s: Done

src/sympc/tensor/replicatedshare_tensor.py
```diff
@@ -1051,6 +1124,73 @@ def method(
...
+        ):
+            raise ValueError("Input tensor to wrap should be a numpy array.")
+        shares = []
+
+        for x_share, y_share in zip(self.shares, other.shares):
+            shares.append(Falcon.wrap2(x_share, y_share))
+
+        config = Config(encoder_base=1, encoder_precision=0)
+        result = ReplicatedSharedTensor(
+            ring_size=2,
```
kamathhrishi, line 1187: Hmmmmm, Why ring size 2?

src/sympc/protocol/falcon/falcon.py
```diff
@@ -553,3 +585,156 @@ def private_compare(x: List[MPCTensor], r: torch.Tensor) -> MPCTensor:
...
+            c (np.ndarray): Input numpy array.
+
+        Returns:
+            result (np.ndarray): Modulo reduction of the exact wrap function.
+
+        Raises:
+            ValueError: If the input values is not a numpy array.
+            ValueError: If signed numpy array are provided as input.
+
+        wrap3 calucaties the carry bit on addition of three values a+b+c(mod 2).
```
kamathhrishi, line 632: Small type :P

src/sympc/protocol/falcon/falcon.py
```diff
@@ -553,3 +585,156 @@ def private_compare(x: List[MPCTensor], r: torch.Tensor) -> MPCTensor:
...
+        """
+        if not isinstance(a, np.ndarray) or not isinstance(b, np.ndarray):
+            raise ValueError("Input value must be a numpy array for wrap2.")
+
+        if a.dtype.kind == "i" or b.dtype.kind == "i":
+            raise ValueError("Wrap2 works only for signed numbers.")
+
+        max_val = np.iinfo(a.dtype).max
+
+        return a > max_val - b
```
kamathhrishi, line 614: How exactly does this comparison operation occur? a and b are owned by 2 different parties, right?

src/sympc/tensor/replicatedshare_tensor.py
rasswanth-s: The `Falcon.wrap2` returns an boolean numpy wrap(1 accounting for wrap , 0 accounting for no wrap)

src/sympc/protocol/falcon/falcon.py
rasswanth-s: fast documentation :smile:

src/sympc/protocol/falcon/falcon.py
rasswanth-s: This computation is done locally in a specific parties side where each is a numpy array.Both a&b belong to the same party.

src/sympc/protocol/aby3/aby3.py
gmuraru: No no, my question was more. From where do we reach a possible `numpy` array here?

rasswanth-s: Oh ok , In Private Compare,we convert torch tensor to numpy. We invoke a multiplication operation ,it reaches here for truncation george.

rasswanth-s: At a high level  ReLU -> Wrap -> PC in the order of invocation . ReLU works with torch tensors,At wrap we convert to Numpy.PC does some additional conversion in Numpy for random generation.I think in Numpy we cannot generate random tensor with custom shape like we do in torch(It has for floats, but not integers),we use torch random tensor genearation , and convert to numpy.

gmuraru: Hmm...but why don't we convert back to tensor in PC?

rasswanth-s: PC is a subroutine in wrap george, I left it like that , as we require numpy there.

gmuraru: yep, but it does not ask my question. but if we convert there back to tensor there, we won't have all the logic where we check if we have a numpy array (in all the other places), right?

gmuraru: ping :eyes:

rasswanth-s: I think not george, for example in the wrap pre processing step `function` we use all the primitive operations of MPC (add,sub,mul),so we had to include support for numpy and a corresponding sanity check for the same.Was this the question, I am not sure , If I answered the question.

rasswanth-s: By all other places  , you meant the (`ABY,addmodprime,submodprime,mulmodprime`) crct george?

gmuraru: So my intention was to remove the numpy logic and convert to ShareTensor/Tensor where we have the numpy array - from my understanding we convert to numpy array only in PC, right (as per comment
"Oh ok , In Private Compare,we convert torch tensor to numpy. We invoke a multiplication operation in the subsequent operation in Private Compare ,it reaches here for truncation george." -- and in this case can't we convert back to tensor in PC such that we won't have all the numpy logic in all the places?)

rasswanth-s: Sorry , My bad, As the comment was placed in ABY3, I was answering in that context.
I will give a comprehensive explanation.
We use `numpy` in wrap context mainly.
Why are we doing that?To suppose the claim, Falcon wrap is designed for `unsigned` numbers , What we really want to do is check for carry using unsigned numbers.

rasswanth-s: http://teaching.idallen.com/dat2343/10f/notes/040_overflow.txt , this article helped me in understanding the subtleties of overflow and underflow

rasswanth-s: Since we use in numpy in wrap and PC is a subroutine of Wrap, it is also made compatible for numpy,

rasswanth-s: The main reason , why we modify all the operations , in many places, is that , when we use numpy in wrap, it undergoes lot of primitive operations of MPC , like the (add, sub, mul) , so we make it compatible with all ops.

rasswanth-s: https://github.com/OpenMined/SyMPC/blob/5b699405716a0e1999b0b84efb8e4a84ca1b7c99/src/sympc/protocol/falcon/falcon.py#L647 I think this function of wrap , would help to understand

gmuraru: I will have a look at the wrap today - I am not sure why we are not able to use the carry bit if we have signed numbers (under the hood - the bit representation should be the same for signed/unsigned numbers -- only that the unsigned numbers have 1 more bit for the representation which for the signed numbers that one is the sign bit)

PR #314, opened 2021-11-02
Pass correct parameters for Conv2d grad

kamathhrishi:
## Description
Previously the parameters passed for conv2d grad did not generalize when these parameters were specified as tuples. They only worked when they were integers. This PR solves the issue when specified as integers or tuples.

### DIFF
--- a/src/sympc/grads/grad_functions.py
+++ b/src/sympc/grads/grad_functions.py
@@ -557,10 +557,10 @@ class GradConv2d(GradFunc):
         output_padding = torch.nn.grad._grad_input_padding(
             grad_output=torch.empty(grad.shape),
             input_size=x.shape,
-            stride=(stride, stride),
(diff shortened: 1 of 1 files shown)


kamathhrishi:
@gmuraru @rasswanth-s Help with failing tests?

src/sympc/grads/grad_functions.py
```diff
@@ -554,13 +554,24 @@ def backward(ctx: Dict[str, Any], grad: MPCTensor) -> Tuple[MPCTensor]:
         out_channels = grad.shape[1]
         min_batch = x.shape[0]
 
+        if type(stride) == int:
```
gmuraru, line 557: MAYBE: Use `isinstance`

gmuraru:
Any update on this?

Issue #324, opened 2022-04-28
How to concatenate `ReplicatedSharedTensor`?


Timo9Madrid7:
Hi,

I noticed `sympc.tensor.static.cat` can concatenate `ShareTensor`s, but it couldn't support `ReplicatedSharedTensor`. So, I wonder if there exists a method to concatenate `ReplicatedSharedTensor`?

Thanks,
Timo

Timo9Madrid7:
I had solved this issue, so I closed it. If anyone meets the same problem, try to do the followings:
1. Create `n` replicated shares.
2. Before sharing these `n` shares,  use `torch.cat` to concatenate shares `shares[i].get_shares()[j]` that belongs to the identical party `j`. 
3. Use `ReplicatedSharedTensor.distribute_shares` or `MPCTensor` to send those shares to each party.

PR #326, opened 2022-04-30
`cat` concatenates `ReplicatedSharedTensor`

Timo9Madrid7:
## Description
- refer to issue #324
- `cat` can concatenates `ShareTensor`, but it is not workable for `ReplicatedSharedTensor`.
- added `cat_replicatedShare_tensor` in *apy.py* and *static.py* for the concatenation of `ReplicatedSharedTensor`.

## Checklist

### DIFF
--- a/src/sympc/tensor/static.py
+++ b/src/sympc/tensor/static.py
@@ -15,12 +15,12 @@ from uuid import UUID
 import numpy as np
 import torch
 
+import sympc.protocol as protocol
 from sympc.session import get_session
 from sympc.tensor.mpc_tensor import MPCTensor
-from sympc.tensor.share_tensor import ShareTensor
 from sympc.tensor.replicatedshare_tensor import ReplicatedSharedTensor
+from sympc.tensor.share_tensor import ShareTensor
 from sympc.utils import parallel_execution
-import sympc.protocol as protocol
 
 
 def stack(tensors: List, dim: int = 0) -> MPCTensor:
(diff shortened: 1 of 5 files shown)


src/sympc/tensor/static.py
```diff
@@ -117,6 +121,27 @@ def cat_share_tensor(session_uuid_str: str, *shares: Tuple[ShareTensor]) -> Shar
...
+        shares (Tuple[ShareTensor]): Replicated shares of the tensors to be concatenated.
+
+    Returns:
+        ReplicatedSharedTensor: Respective replicated shares after concatenation
+    """
+    
+    session = get_session(session_uuid_str)
+    result = ReplicatedSharedTensor(session_uuid=UUID(session_uuid_str), config=session.config)
+    
+    cat_result = [torch.tensor([]).type(torch.LongTensor) for _ in range(len(shares[0].shares))]
```
gmuraru, line 138: MAYBE: Could you save the `len(shares[0].shares)` in a variable and then use it here and on line 140?

gmuraru:
Only one comment! I think you also need to run the `black` formatted and then we are ready to merge this :rocket:

src/sympc/tensor/static.py
Timo9Madrid7: Yes, nice catch, that will more efficient. 👍

Timo9Madrid7:
> Only one comment! I think you also need to run the `black` formatted and then we are ready to merge this 🚀

Ok, I have applied `black` formatter for `api.py` and `tensor/static.py`.

gmuraru:
Ah..you also need to use the `isort`.
The better solution is to install the `pre-commit` hooks and they should do the job!

Timo9Madrid7:
> Ah..you also need to use the `isort`. The better solution is to install the `pre-commit` hooks and they should do the job!

Alright! These two files have been checked and passed those tests. Thank you for guiding me.
```
$ pre-commit run --files src/sympc/api.py src/sympc/tensor/static.py
```

gmuraru:
> > Ah..you also need to use the `isort`. The better solution is to install the `pre-commit` hooks and they should do the job!
> 
> Alright! These two files have been checked and passed those tests. Thank you for guiding me.
> 
> ```
> $ pre-commit run --files src/sympc/api.py src/sympc/tensor/static.py
> ```

No worries!

gmuraru (approved):
LGTM!

